Issue #18, opened 2023-08-14
support r1cs


mimoo:
> in general, it would be cool to be able to choose r1cs as a backend (via some --r1cs flag) and noname would dump r1cs constraints/witness instead (I think it shouldn't be too hard, as in: the public input should have the value 1 be the first value, the witness symbolic table would have a single column (z), constraints would add to matrices a, b, c instead of adding plonk custom gates to the gates array

from https://github.com/zksecurity/noname/issues/17

PR #30, opened 2024-03-28
WIP: Modularize backends

katat:
for #18 

- [x] Refactor with ark_ff::Field
- [x] Modularize backends
- [x] Existing tests passed
- [ ] Add R1CS backend


katat:
The changes from the actual pallas field type to trait `ark_ff::Field` propagate to the variables such as:

- BUILTIN_FNS https://github.com/katat/noname/blob/aa9687a6439a2477d8ceee1d3f797f92ad7000aa/src/imports.rs#L77
- CRYPTO_MODULES https://github.com/katat/noname/blob/aa9687a6439a2477d8ceee1d3f797f92ad7000aa/src/stdlib/mod.rs#L24
- CRYPTO_FNS https://github.com/katat/noname/blob/aa9687a6439a2477d8ceee1d3f797f92ad7000aa/src/stdlib/crypto.rs#L17

These variables are `const` or `static`, so they won't be able to support using generic type without make some complex tricks using keywords such as `Box` or `dyn`.

I am wondering if it makes more sense to have the NAST struct to encapsulate these variables, so they can inherit the generic type from their container.

src/var.rs
```diff
@@ -69,13 +76,13 @@ pub enum Value {
...
-impl From<Field> for Value {
-    fn from(field: Field) -> Self {
-        Self::Constant(field)
-    }
-}
+// impl<B: Backend> From<B> for Value<B> {
+//     fn from(field: B) -> Self {
+//         Self::Constant(field)
+//     }
+// }
```
mimoo, line 83: remove this code if it wasn't needed :o

src/witness.rs
```diff
@@ -1,45 +1,51 @@
...
 #[derive(Debug)]
-pub struct Witness(Vec<[Field; NUM_REGISTERS]>);
+pub struct Witness<F: Field>(Vec<[F; NUM_REGISTERS]>);
 
-impl Witness {
+impl<F: Field + helpers::PrettyField> Witness<F> {
     /// kimchi uses a transposed witness
-    pub fn to_kimchi_witness(&self) -> [Vec<Field>; NUM_REGISTERS] {
-        let transposed = vec![Vec::with_capacity(self.0.len()); NUM_REGISTERS];
+    pub fn to_kimchi_witness(&self) -> [Vec<F>; NUM_REGISTERS] {
```
mimoo, line 45: sounds like this should be moved to the kimchi backend module as well

src/witness.rs
```diff
@@ -1,45 +1,51 @@
...
+        let transposed = (0..NUM_REGISTERS)
+            .map(|_| Vec::with_capacity(self.0.len()))
+            .collect::<Vec<_>>();
```
mimoo, line 48: why this change?

src/compiler.rs
```diff
@@ -143,20 +149,26 @@ pub fn get_nast(
...
+    tast: TypeChecker<B>,
+    backend: B,
+) -> miette::Result<CompiledCircuit<B>> {
+    CircuitWriter::generate_circuit(tast, backend).into_miette(sources)
+}
+
+pub struct GeneratedWitness<B: Backend> {
+    pub all_witness: Witness<B::Field>,
+    pub full_public_inputs: Vec<B::Field>,
+    pub public_outputs: Vec<B::Field>,
```
mimoo, line 163: need documentation

mimoo: * all_witness is the entire witness vector, comprising the public input and the private input
* full_public_inputs is the public input part of the previous vector, including the public output at the end
* public_outputs is only the public output part of the previous vector

to be honest I don't remember why I returned that :)

mimoo: (seems kinda inefficient)

src/witness.rs
```diff
@@ -196,12 +201,15 @@ impl CompiledCircuit {
         // compute each rows' vars, except for the deferred ones (public output)
```
mimoo, line 201: so I think what comes after should be part of the backend as well, this is very specific to kimchi/plonk

mimoo: so perhaps the backend should have a `fn generate_witness()` which we could call at this point

src/stdlib/mod.rs
```diff
@@ -81,12 +55,58 @@ pub const QUALIFIED_BUILTINS: &str = "std/builtins";
...
+pub fn builtin_fns<B: Backend>() -> Vec<FnInfo<B>> {
+    BUILTIN_FNS_SIGS
+        .iter()
+        .map(|(sig, _)| get_builtin_fn::<B>(sig))
+        .collect()
+}
+
+pub fn has_builtin_fn(name: &str) -> bool {
+    BUILTIN_FNS_SIGS.iter().any(|(_, s)| s.name.value == name)
+}
```
mimoo, line 103: can you create a PR that we could merge first that refactors the builtins? This would simplify this PR :o

mimoo (commented):
First of all this looks great, I think this is a great step towards what we want to achieve. I've left a number of comments, and I will continue to do so, but I wanted to get the big ones now rather than later. To summarize:

1. can you create a separate PR that we can merge first for the refactor of the builtins?
2. there's a number of changes that are unrelated to this PR, could you also move them to another PR? This would make reviewing this PR easier as I can focus on the main changes :)
3. I've noted a number of places where I think things should be moved to the backend, let me know if these make sense, otherwise we can have a convo on that :o

src/cli/cmd_build_and_check.rs
```diff
@@ -104,22 +102,23 @@ pub fn cmd_check(args: CmdCheck) -> miette::Result<()> {
         .path
         .unwrap_or_else(|| std::env::current_dir().unwrap().try_into().unwrap());
 
+    // TODO: field type should be determined by a flag
     // produce all TASTs and stop here
-    produce_all_asts(&curr_dir)?;
+    produce_all_asts::<Kimchi>(&curr_dir)?;
```
mimoo, line 107: I think this is the right approach, we can have another PR later to introduce different choices of backend 👍

src/backends/kimchi/fp_kimchi.rs
```diff
@@ -0,0 +1,829 @@
+use ark_ff::Field;
+use kimchi::mina_poseidon::poseidon::ArithmeticSpongeParams;
+
+use std::str::FromStr;
+
+pub fn params<F>() -> ArithmeticSpongeParams<F>
```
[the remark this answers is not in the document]
katat: This is used by the generic [poseidon function](https://github.com/iden3/circom/blob/1322ba6e0289b86f97414eeb67723476c4aea1f4/dag/src/r1cs_porting.rs#L15), that forces this params function to have generic field.

So looks like this poseidon builtin function should be implemented for each backend, then we can just import the params from the kimchi.

src/circuit_writer/writer.rs
```diff
@@ -102,7 +106,7 @@ impl PartialEq for AnnotatedCell {
 
 impl PartialOrd for AnnotatedCell {
     fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
-        self.cell.partial_cmp(&other.cell)
+        Some(self.cmp(other))
```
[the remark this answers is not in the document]
katat: it is a clippy suggestion :)

src/var.rs
```diff
@@ -32,29 +33,35 @@ impl CellVar {
 }
 
 /// The signature of a hint function
-pub type HintFn = dyn Fn(&CompiledCircuit, &mut WitnessEnv) -> Result<Field>;
+pub type HintFn<B> = dyn Fn(
```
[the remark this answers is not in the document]
katat: It can. But somehow clippy complains.

src/backends/kimchi/fp_kimchi.rs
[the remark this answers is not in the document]
mimoo: my question was, why not import https://github.com/o1-labs/proof-systems/blob/master/poseidon/src/pasta/fp_kimchi.rs#L9

but I guess the problem here is that this crate is not published on crates.io :D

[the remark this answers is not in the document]
mimoo: oh and yeah this should not have a generic parameter here, it should be specific for a backend that has a specific field.

katat:
The other PR with same purpose was merged.

PR #33, opened 2024-04-11
refactor: builtins

katat:
The static and const builtins doesn't work for generic types. These changes are for adapting to different backends.

### DIFF
--- a/src/stdlib/mod.rs
+++ b/src/stdlib/mod.rs
@@ -21,48 +21,6 @@ use self::crypto::get_crypto_fn;
 
 pub mod crypto;
 
-pub fn get_std_fn(submodule: &str, fn_name: &str, span: Span) -> Result<FnInfo> {
-    match submodule {
-        "crypto" => get_crypto_fn(fn_name)
-            .ok_or_else(|| {
-                Error::new(
-                    "type-checker",
-                    ErrorKind::UnknownExternalFn(submodule.to_string(), fn_name.to_string()),
-                    span,
-                )
-            }),
-        _ => Err(Error::new(
-            "type-checker",
-            ErrorKind::StdImport(submodule.to_string()),
-            span,
-        )),
-    }
-}
-
-/// Takes a list of function signatures (as strings) and their associated function pointer,
-/// returns the same list but with the parsed functions (as [FunctionSig]).
-pub fn parse_fn_sigs(fn_sigs: &[(&str, FnHandle)]) -> HashMap<String, FnInfo> {
-    let mut functions = HashMap::new();
-    let ctx = &mut ParserCtx::default();
-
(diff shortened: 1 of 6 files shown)


src/stdlib/mod.rs
```diff
@@ -17,21 +17,13 @@ use crate::{
     var::{ConstOrCell, Var},
 };
 
-use self::crypto::CRYPTO_FNS;
+use self::crypto::get_crypto_fn;
 
 pub mod crypto;
 
-pub static CRYPTO_MODULE: Lazy<BuiltinModule> = Lazy::new(|| {
-    let functions = parse_fn_sigs(&CRYPTO_FNS);
```
mimoo, line 25: can you delete the `parse_fn_sigs` function now?

src/type_checker/mod.rs
```diff
@@ -91,12 +91,7 @@ impl TypeChecker {
     }
 
     pub(crate) fn fn_info(&self, qualified: &FullyQualified) -> Option<&FnInfo> {
-        if qualified.module == Some(UserRepo::new("std/builtins")) {
-            // if it's a built-in: get it from a global
-            BUILTIN_FNS.get(&qualified.name)
-        } else {
-            self.functions.get(qualified)
-        }
```
mimoo, line 99: I'm puzzled as to why this code existed before, I guess it was useless right?

src/type_checker/mod.rs
```diff
@@ -154,8 +149,8 @@ impl TypeChecker {
 
         // initialize it with the standard library
         let crypto_module = ModulePath::Absolute(UserRepo::new("std/crypto"));
-        for (fn_name, fn_info) in CRYPTO_MODULE.functions.iter() {
-            let qualified = FullyQualified::new(&crypto_module, fn_name);
+        for fn_info in crypto_fns().iter() {
```
mimoo, line 152: btw you should be able to write `for fn_info in crypto_fns()` no?

mimoo (approved):
Nice! Looks good to me!

src/stdlib/mod.rs
```diff
@@ -17,21 +17,13 @@ use crate::{
...
-    BuiltinModule { functions }
-});
-
 pub fn get_std_fn(submodule: &str, fn_name: &str, span: Span) -> Result<FnInfo> {
```
mimoo, line 24: is this function used anywhere?

katat: it seems no. may be dead code?

src/type_checker/mod.rs
katat: My understanding is it was trying to differentiate the crypto builtin functions and native functions.
But since both the builtins and natives are stored in the same place now, this code seem deprecated.

PR #48, opened 2024-04-22
Refactor for basic constraints

katat:
- [x] add basic constraint interfaces to the backend: `neg` / `add` / `mul` / `assert_eq`
- [x] refactor boolean constraints 
- [x] refactor field constraints 
- [x] refactor builtin constraints
- [x] move some logic from add_public_outputs and add_public_inputs to the backend (to remove the add_gate stuff)


src/backends/mod.rs
```diff
@@ -35,6 +35,29 @@ pub trait Backend: Clone {
...
+        span: Span,
+    ) -> ConstOrCell<Self::Field>;
+
+    /// basic constraint multiplication
+    fn enforce_mul_constraint(
+        &mut self,
+        lhs: &ConstOrCell<Self::Field>,
+        rhs: &ConstOrCell<Self::Field>,
+        span: Span,
+    ) -> ConstOrCell<Self::Field>;
```
mimoo, line 59: I would call these `neg`, `add`, and `mul` as they might not create a constraint eventually.

src/constraints/boolean.rs
```diff
@@ -11,23 +11,22 @@ use crate::{
...
-        ConstOrCell::Cell(var) => compiler.backend.add_generic_gate(
-            "constraint to validate a boolean (`x(x-1) = 0`)",
-            // x^2 - x = 0
-            vec![Some(*var), Some(*var), None],
-            vec![one.neg(), zero, zero, one],
-            span,
-        ),
+        ConstOrCell::Cell(_) => {
+            // x * (x - 1)
+            let x_1 = field::sub(compiler, xx, &ConstOrCell::Const(one.neg()), span);
```
mimoo, line 27: `one` not `one.neg()`

src/constraints/field.rs
```diff
@@ -326,55 +170,23 @@ fn equal_cells<B: Backend>(
...
-            compiler.backend.add_generic_gate(
-                "constraint #3 for the equals gadget (res * diff = 0)",
-                vec![Some(res), Some(diff)],
-                // res * diff = 0
-                vec![zero, zero, zero, one],
-                span,
-            );
+            let res_mul_diff = mul(compiler, &ConstOrCell::Cell(res), &diff, span)[0];
+            // ensure that res * diff = 0
+            sub(compiler, &res_mul_diff, &ConstOrCell::Const(zero), span);
```
mimoo, line 181: this is a no-op, instead what we want to do here is `assert_eq(res_mul_diff, zero)`, where `assert_eq` is also exposed by the backend

src/constraints/field.rs
```diff
@@ -326,55 +170,23 @@ fn equal_cells<B: Backend>(
...
 
-            compiler.backend.add_generic_gate(
-                "constraint #4 for the equals gadget (diff_inv * diff = one_minus_res)",
-                vec![Some(diff_inv), Some(diff), Some(one_minus_res)],
-                // diff_inv * diff - one_minus_res = 0
-                vec![zero, zero, one.neg(), one],
-                span,
-            );
+            let diff_inv_mul_diff = mul(compiler, &ConstOrCell::Cell(diff_inv), &diff, span)[0];
+            sub(compiler, &diff_inv_mul_diff, &one_minus_res, span);
```
mimoo, line 189: same here, either `assert_eq(diff_inv_mul_diff, one_minus_res)` or assert that the result of `sub` is equal to zero.

BTW we also should add [`#[must_use]`](https://doc.rust-lang.org/reference/attributes/diagnostics.html#the-must_use-attribute) to these functions (sub, neg, etc.) as well as trait functions (the one you added to do sub, neg, mul, etc.) so that the compiler yells at us if we don't use a var or a cvar from an operation

src/constraints/field.rs
```diff
@@ -11,65 +12,30 @@ use ark_ff::{One, Zero};
...
     lhs: &ConstOrCell<B::Field>,
     rhs: &ConstOrCell<B::Field>,
     span: Span,
 ) -> Var<B::Field> {
-    let zero = B::Field::zero();
-    let one = B::Field::one();
-
-    match (lhs, rhs) {
-        // 2 constants
-        (ConstOrCell::Const(lhs), ConstOrCell::Const(rhs)) => Var::new_constant(*lhs + *rhs, span),
```
mimoo, line 26: so my feeling is that it would be more elegant to leave the code handling constants here. Perhaps this means that in the backend we would also have: `add_const(const: F, cvar: CellVar<F>)` and `scale(coeff: F, cvar: CellVar<F>)`

this way the handling of pure constants is not done in the backend

examples/arithmetic.no
```diff
@@ -1,4 +1,7 @@
 fn main(pub public_input: Field, private_input: Field) {
     let xx = private_input + public_input;
     assert_eq(xx, 2);
+
+    let yy = private_input - public_input;
+    assert_eq(yy, 0);
```
mimoo, line 6: probably good to have this change in a different PR :)

src/backends/mod.rs
mimoo: also as I hint later, I think they should not receive `ConstOrCell` but `CellVar` instead

mimoo (commented):
I think this is going in the right direction! Obviously this is going to make our constraints less efficient, but we can fix this in a later PR

src/backends/mod.rs
katat: It now uses CellVar at the backend layer

src/constraints/boolean.rs
katat: fixed

src/constraints/field.rs
katat: it is now enforced by `assert_eq`

src/constraints/field.rs
katat: this one should be fixed

src/backends/kimchi/mod.rs
```diff
@@ -133,58 +136,11 @@ impl KimchiVesta {
             pending_generic_gate: None,
             debug_info: vec![],
             finalized: false,
+            public_input_size: 0,
```
mimoo, line 139: would it make sense to pass that as argument in `new()`?

src/backends/kimchi/mod.rs
```diff
@@ -546,4 +551,159 @@ impl Backend for KimchiVesta {
 
         res
     }
+
+    fn constraint_neg(&mut self, var: &CellVar, span: Span) -> CellVar {
```
mimoo, line 555: remove `constraint` from the name (so just `neg`) as it wrongly imply that we will create a constraint here (same for others)

src/backends/mod.rs
```diff
@@ -35,6 +35,27 @@ pub trait Backend: Clone {
...
+    fn constraint_mul(&mut self, lhs: &CellVar, rhs: &CellVar, span: Span) -> CellVar;
+
+    /// add a constraint to assert a var is multiplied by a constant
+    fn constraint_mul_const(&mut self, var: &CellVar, cst: &Self::Field, span: Span) -> CellVar;
+
+    /// add a constraint to assert a var equals a constant
+    fn constraint_eq_const(&mut self, var: &CellVar, cst: Self::Field, span: Span);
+
+    /// add a constraint to assert a var equals another var
+    fn constraint_eq_var(&mut self, lhs: &CellVar, rhs: &CellVar, span: Span);
```
mimoo, line 57: none of these functions have to add a constraint, so constraint should be removed from the name and the doc

mimoo (commented):
went through it quickly and left you a number of comments, but the main one will be the same as with the other PR :D can you create easy-to-review commits?

src/backends/kimchi/mod.rs
mimoo: actually it doesn't make sense, rereading things, as it gets calculated when we call add_public_input and add_public_output

katat: yeah, that was wondering this fact

PR #52, opened 2024-04-26
Support R1CS

katat:
## snarkjs testing workflow

Here are the commands to verify the whole process.

```bash
# create new noname package. this will create a new folder test, containing the noname code.
cargo run new -p test

# trigger the noname run subcommand to generate .r1cs and .wtns binary files
cargo run run --backend r1cs --path test --private-inputs '{"yy": "1"}' --public-inputs '{"xx": "2"}'

# check the consistency between the witness and r1cs constraints via snarkjs
snarkjs wchk test/output.r1cs test/output.wtns


# for generating a proof and verify with it via snarkjs
snarkjs powersoftau new bls12-381 12 pot12_0000.ptau -v
snarkjs powersoftau contribute pot12_0000.ptau pot12_0001.ptau --name="First contribution" -v
snarkjs powersoftau prepare phase2 pot12_0001.ptau pot12_final.ptau -v
snarkjs groth16 setup test/output.r1cs pot12_final.ptau test_0000.zkey
snarkjs zkey contribute test_0000.zkey test_0001.zkey --name="1st Contributor Name" -v
snarkjs zkey export verificationkey test_0001.zkey verification_key.json
snarkjs groth16 prove test_0001.zkey test/output.wtns proof.json public.json
snarkjs groth16 verify verification_key.json public.json proof.json


# at the end, it should display "snarkJS: OK!"

```

## Missing meta data
 - It currently doesn't store the number of outputs and private inputs to the r1cs file.
 - Maybe we can shift and have the backend store the public outputs and private inputs? 
 - [x] Either we can pass these meta data from the circuit writer or from within backend. The latter approach seems cleaner.

## witness data arrangement
 - snarkjs expect the public inputs are arranged at first. This is crucial for generating the correct proving index.
 - [x] Make sure the public inputs are always arranged at first (relating to how the new vars are recorded in `witness_vars`)

## constraint public inputs / outputs
- in kimchi backend, it needs to add constraints that will be cancelled by the verifier.
- [x] research: it seems we don't need to do this in r1cs as long as the public inputs can be correctly recognized from the witness data

## Implement finalize_circuit interface
 - [x] what do we need to check before finalizing it?

### DIFF
--- a/src/backends/mod.rs
+++ b/src/backends/mod.rs
@@ -4,17 +4,23 @@ use ark_ff::{Field, Zero};
 use num_bigint::BigUint;
 
 use crate::{
-    circuit_writer::{DebugInfo, GateKind},
     compiler::Sources,
     constants::Span,
     error::{Error, ErrorKind, Result},
-    helpers::PrettyField,
     imports::FnHandle,
     var::{CellVar, Value, Var},
     witness::WitnessEnv,
 };
 
+use self::{kimchi::KimchiVesta, r1cs::R1csBls12_381};
+
 pub mod kimchi;
+pub mod r1cs;
+
+pub enum BackendKind {
+    KimchiVesta(KimchiVesta),
+    R1CS(R1csBls12_381),
+}
 
 // TODO: should it be cloneable? It is now so because FnInfo needs to be cloneable.
 pub trait Backend: Clone {
--- a/src/backends/mod.rs
+++ b/src/backends/mod.rs
@@ -86,7 +86,12 @@ pub trait Backend: Clone {
         self.compute_val(env, self.witness_vars(var), var.index)
     }
 
-    fn compute_val(&self, env: &mut WitnessEnv<Self::Field>, val: &Value<Self>, var_index: usize) -> Result<Self::Field> {
+    fn compute_val(
+        &self,
+        env: &mut WitnessEnv<Self::Field>,
+        val: &Value<Self>,
+        var_index: usize,
+    ) -> Result<Self::Field> {
         if let Some(res) = env.cached_values.get(&var_index) {
             return Ok(*res);
         }
--- a/src/backends/mod.rs
+++ b/src/backends/mod.rs
@@ -1,6 +1,7 @@
-use std::collections::HashMap;
+use std::{collections::HashMap, str::FromStr};
 
 use ark_ff::{Field, Zero};
+use num_bigint::BigUint;
 
 use crate::{
     circuit_writer::{DebugInfo, GateKind},
@@ -18,7 +19,7 @@ pub mod kimchi;
 // TODO: should it be cloneable? It is now so because FnInfo needs to be cloneable.
 pub trait Backend: Clone {
     /// The circuit field / scalar field that the circuit is written on.
-    type Field: Field + PrettyField;
+    type Field: Field + FromStr + TryFrom<BigUint> + TryInto<BigUint> + Into<BigUint>;
 
     /// The generated witness type for the backend. Each backend may define its own witness format to be generated.
     type GeneratedWitness;
--- a/src/backends/mod.rs
+++ b/src/backends/mod.rs
@@ -32,7 +32,7 @@ pub trait Backend: Clone {
 
     /// This provides a standard way to access to all the internal vars.
     /// Different backends should be accessible in the same way by the variable index.
-    fn witness_vars(&self) -> &HashMap<usize, Value<Self>>;
+    fn witness_vars(&self, var: CellVar) -> &Value<Self>;
 
     // TODO: as the builtins grows, we might better change this to a crypto struct that holds all the builtin function pointers.
     /// poseidon crypto builtin function for different backends
@@ -89,7 +89,7 @@ pub trait Backend: Clone {
             return Ok(*res);
         }
 
-        match &self.witness_vars()[&var.index] {
+        match &self.witness_vars(var) {
             Value::Hint(func) => {
                 let res = func(self, env)
                     .expect("that function doesn't return a var (type checker error)");
--- a/src/backends/mod.rs
+++ b/src/backends/mod.rs
@@ -83,17 +83,19 @@ pub trait Backend: Clone {
     /// - The symbolic variables are stored in the witness_vars.
     /// - The computed values are stored in the cached_values.
     fn compute_var(&self, env: &mut WitnessEnv<Self::Field>, var: CellVar) -> Result<Self::Field> {
-        // fetch cache first
-        // TODO: if self was &mut, then we could use a Value::Cached(Field) to store things instead of that
-        if let Some(res) = env.cached_values.get(&var) {
+        self.compute_val(env, self.witness_vars(var), var.index)
+    }
+
+    fn compute_val(&self, env: &mut WitnessEnv<Self::Field>, val: &Value<Self>, var_index: usize) -> Result<Self::Field> {
+        if let Some(res) = env.cached_values.get(&var_index) {
             return Ok(*res);
         }
 
-        match &self.witness_vars(var) {
+        match &val {
             Value::Hint(func) => {
                 let res = func(self, env)
                     .expect("that function doesn't return a var (type checker error)");
-                env.cached_values.insert(var, res);
+                env.cached_values.insert(var_index, res);
                 Ok(res)
             }
             Value::Constant(c) => Ok(*c),
@@ -102,20 +104,20 @@ pub trait Backend: Clone {
                 for (coeff, var) in lc {
                     res += self.compute_var(env, *var)? * *coeff;
                 }
-                env.cached_values.insert(var, res); // cache
+                env.cached_values.insert(var_index, res); // cache
                 Ok(res)
             }
             Value::Mul(lhs, rhs) => {
                 let lhs = self.compute_var(env, *lhs)?;
                 let rhs = self.compute_var(env, *rhs)?;
                 let res = lhs * rhs;
-                env.cached_values.insert(var, res); // cache
+                env.cached_values.insert(var_index, res); // cache
                 Ok(res)
             }
             Value::Inverse(v) => {
                 let v = self.compute_var(env, *v)?;
                 let res = v.inverse().unwrap_or_else(Self::Field::zero);
-                env.cached_values.insert(var, res); // cache
+                env.cached_values.insert(var_index, res); // cache
                 Ok(res)
             }
             Value::External(name, idx) => Ok(env.get_external(name)[*idx]),
--- a/src/backends/r1cs/mod.rs
+++ b/src/backends/r1cs/mod.rs
@@ -16,14 +16,14 @@ use self::builtin::poseidon;
 use super::Backend;
 
 /// Linear combination of variables and constants
-/// For example, the linear combination is represented as a * f_a + b * f_b + c
+/// For example, the linear combination is represented as a * f_a + b * f_b + f_c
 /// f_a and f_b are the coefficients of a and b respectively.
 /// a and b are represented by CellVar
-/// the constant c is represented by the constant field, will always multiply with the variable at index 0 which is always 1
+/// the constant f_c is represented by the constant field, will always multiply with the variable at index 0 which is always 1
 #[derive(Clone, Debug)]
 pub struct LinearCombination {
-    pub terms: Option<HashMap<CellVar, Fr>>,
-    pub constant: Option<Fr>,
+    pub terms: HashMap<CellVar, Fr>,
+    pub constant: Fr,
 }
 
 impl LinearCombination {
@@ -31,15 +31,11 @@ impl LinearCombination {
     pub fn evaluate(&self, witness: &HashMap<usize, Fr>) -> Fr {
         let mut sum = Fr::zero();
 
-        if let Some(terms) = &self.terms {
-            for (var, factor) in terms {
-                sum += *witness.get(&var.index).unwrap() * factor;
-            }
+        for (var, factor) in &self.terms {
+            sum += *witness.get(&var.index).unwrap() * factor;
         }
 
-        if let Some(constant) = &self.constant {
-            sum += *constant;
-        }
+        sum += &self.constant;
 
         sum
     }
@@ -206,10 +202,8 @@ impl Backend for R1csBls12_381 {
         let mut written_vars = HashSet::new();
         for constraint in &self.constraints {
             for lc in constraint.as_array() {
-                if let Some(terms) = &lc.terms {
-                    for var in terms.keys() {
-                        written_vars.insert(var.index);
-                    }
+                for var in lc.terms.keys() {
+                    written_vars.insert(var.index);
                 }
             }
         }
@@ -243,14 +237,12 @@ impl Backend for R1csBls12_381 {
 
         for constraint in &self.constraints {
             for lc in &constraint.as_array() {
-                if let Some(terms) = &lc.terms {
-                    for var in terms.keys() {
-                        if witness.contains_key(&var.index) {
-                            continue;
-                        }
-                        let val = self.compute_var(witness_env, *var)?;
-                        witness.insert(var.index, val);
+                for var in lc.terms.keys() {
+                    if witness.contains_key(&var.index) {
+                        continue;
                     }
+                    let val = self.compute_var(witness_env, *var)?;
+                    witness.insert(var.index, val);
                 }
             }
             // assert a * b = c
@@ -278,7 +270,7 @@ impl Backend for R1csBls12_381 {
         todo!()
     }
     
-    /// to constraint:
+    /// To constrain:
     /// x + (-x) = 0
     /// given:
     /// a * b = c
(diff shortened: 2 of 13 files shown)


src/backends/r1cs/mod.rs
```diff
@@ -0,0 +1,557 @@
...
+    fn new_internal_var(
+        &mut self,
+        val: crate::var::Value<Self>,
+        span: crate::constants::Span,
+    ) -> CellVar {
+        let var = CellVar::new(self.next_variable, span);
+        self.next_variable += 1;
+
+        // store it in the circuit_writer
+        self.witness_vars.insert(var.index, val);
```
mimoo, line 161: the way I see it you could rewrite this as

```rust
let var = CellVar::new(self.witness_vars.len(), span);
self.witness_vars.push(val);
var
```

src/backends/r1cs/mod.rs
```diff
@@ -0,0 +1,557 @@
...
+        &mut self,
+        //todo: do we need this?
+        label: Option<&'static str>,
+        value: Self::Field,
+        span: crate::constants::Span,
+    ) -> CellVar {
+        let x = self.new_internal_var(Value::Constant(value), span);
+        self.assert_eq_const(&x, value, span);
+
+        x
```
mimoo, line 176: replace with `CellVar::new_constant` for now

src/backends/r1cs/mod.rs
```diff
@@ -0,0 +1,557 @@
...
+        // store the return value in the public input that was created for that
+        if let Some(public_output) = public_output {
+            let cvars = &public_output.cvars;
+
+            for (pub_var, ret_var) in cvars.clone().iter().zip(returned_cells.unwrap()) {
+                // replace the computation of the public output vars with the actual variables being returned here
+                let var_idx = pub_var.idx().unwrap();
+                let prev = self
+                    .witness_vars
+                    .insert(var_idx, Value::PublicOutput(Some(ret_var)));
```
mimoo, line 200: the way I would do it is just do `self.witness_vars[var_idx] = Value::PublicOutput(Some(ret_var));

src/backends/r1cs/mod.rs
```diff
@@ -0,0 +1,557 @@
...
+                assert!(prev.is_some());
```
mimoo, line 201: maybe `assert!(matches!(prev, Some(Value::PublicOutput(None)))` might be nicer

src/backends/r1cs/mod.rs
```diff
@@ -0,0 +1,557 @@
...
+            }
+        }
```
mimoo, line 203: now that I'm re-reading this I'm wondering how this (and kimchi) are constraining the public outputs... we should have `compiler.assert_eq()` between the `CellVar` returned by `compile_main_function` and the ones in the public input/output...

src/backends/r1cs/mod.rs
```diff
@@ -0,0 +1,557 @@
...
+            "eq constraint: lhs = rhs",
+            Constraint{a, b, c},
+            span
+        );
+    }
+    
+    /// Adds the public input cell vars
+    fn add_public_input(&mut self, val: Value<Self>, span: crate::constants::Span) -> CellVar {
+        let var = self.new_internal_var(val, span);
+        self.public_inputs.push(var);
```
mimoo, line 544: wondering why we track these

src/backends/r1cs/mod.rs
```diff
@@ -0,0 +1,557 @@
...
+        };
+
+        self.add_constraint(
+            "eq constraint: x = cst",
+            Constraint{a, b, c},
+            span
+        );
+    }
+
+    /// to constraint:
```
mimoo, line 508: to constrain*

src/backends/r1cs/mod.rs
```diff
@@ -0,0 +1,557 @@
...
+    /// lhs = rhs
+    /// given:
+    /// a * b = c
+    /// then:
+    /// a = lhs
+    /// b = 1
+    /// c = rhs
```
mimoo, line 515: also I think these kind of comments should be better as just comments in the implementation rather than in the rust doc comment, also you don't usually write rust doc comment on a trait implementation

src/backends/r1cs/mod.rs
```diff
@@ -0,0 +1,557 @@
...
+    fn assert_eq_var(&mut self, lhs: &CellVar, rhs: &CellVar, span: crate::constants::Span) {
+        let one = Fr::from(1);
+
+        let a = LinearCombination {
+            terms: Some(HashMap::from_iter(vec![(*lhs, one)])),
```
mimoo, line 520: `&[*lhs, one]` should work no?

src/backends/r1cs/mod.rs
```diff
@@ -0,0 +1,557 @@
...
+            constant: None,
+        };
+
+        let b = LinearCombination {
+            terms: None,
+            constant: Some(one),
+        };
```
mimoo, line 527: `LinearCombination::one()` should be cleaner :o

src/backends/r1cs/mod.rs
```diff
@@ -0,0 +1,557 @@
...
+
+        let b = LinearCombination {
+            terms: Some(HashMap::from_iter(vec![(*rhs, one)])),
+            constant: None,
+        };
+
+        let c = LinearCombination {
+            terms: Some(HashMap::from_iter(vec![(res, one)])),
+            constant: None,
+        };
```
mimoo, line 427: `LinearCombination::from_var(res)` would be cleaner

src/backends/r1cs/mod.rs
```diff
@@ -0,0 +1,557 @@
...
+                if let Some(terms) = &lc.terms {
+                    for var in terms.keys() {
+                        if witness.contains_key(&var.index) {
+                            continue;
+                        }
+                        let val = self.compute_var(witness_env, *var)?;
+                        witness.insert(var.index, val);
+                    }
+                }
+            }
```
mimoo, line 255: I'm not sure I follow why we do this here. Why can't we just do something like

```rust
let witness = witness.iter().map(|v| self.compute_val(witness_env, v)).collect_vec();
```

although:

* you'd have to skip the public output and compute it at the end
* you'd have to have a `compute_val` function that implements this logic:

Screenshot 2024-05-01 at 11 41 30 AM

mimoo: the whole idea is that:

* the witness vector is a single vector
* the index in that vector can be the same as the index of a CellVar
* when we compile the circuit, we know exactly when we create new variables by pushing in that witness vector
* so we can have a "symbolic" witness vector that tells us how to compute each cell of that vector at runtime

src/cli/cmd_build_and_check.rs
```diff
@@ -295,3 +296,68 @@ pub fn cmd_test(args: CmdTest) -> miette::Result<()> {
...
+    
+    let private_inputs = if let Some(s) = args.private_inputs {
+        parse_inputs(&s)?
+    } else {
+        JsonInputs::default()
+    };
+    
+    match backend.as_str() {
+        "kimchi-vesta" => todo!(),
+        "r1cs" => {
```
mimoo, line 341: maybe `snarkjs-r1cs`? As it has a particular serialization

src/cli/cmd_build_and_check.rs
```diff
@@ -295,3 +296,68 @@ pub fn cmd_test(args: CmdTest) -> miette::Result<()> {
...
     Ok(())
 }
+
+#[derive(clap::Parser)]
+pub struct CmdRun {
+    /// noname file to run
+    #[clap(short, long, value_parser)]
+    path: Option<PathBuf>,
+
+    /// backend to use for running the noname file.
```
mimoo, line 306: suggest adding the backends currently supported in the comment

mimoo (commented):
This looks really great! I left you a few comments. But the important ones:

* did we forget to constrain the returned CellVars with the public output CellVars o_O? https://zksecurity.github.io/noname/public-outputs.html doesn't even mention it ;_;
* it'd be nice to document the snarkjs expected serialization for .r1cs and .witness, maybe in the book?
* I don't think we should use a hashmap for the symbolic witness vector, I think we can simplify that a lot

src/backends/r1cs/mod.rs
katat: It is for `SnarkjsExporter` to reorder the inputs in the witness vector.

src/backends/r1cs/mod.rs
katat: I am wondering if we should have the `witness_vars` as `Vec<(CellVar, Value)>`, instead of just `Vec<Value>`.

The `compute_var` require the argument in CellVar type, while the other places require `Value`.

With both stored in the vector, we can do the following without the need to retrieve the cell vars from the linear combinations in constraints.
```
let witness = self.witness_vars.iter().map(|v| self.compute_val(witness_env, v.0)).collect_vec();
```

mimoo: but here we are only interested in the index in CellVar here, which is going to be 0, 1, 2, 3, etc.

katat: and we need to recreated the CellVar with a span in order to pass into the `compute_var`, right?

mimoo: I was thinking that compute_var is overkill here, we can export the match statement inside of it and have it live in another fn

katat: Ah, so we will need to refactor the `compute_var` to call `compute_val` which encapsulate the match statements for `symbolic vars` (the Value type)

katat: https://github.com/zksecurity/noname/issues/60

katat: lmk if this is correct way: 530d2b47a2b59b3e26dca2ed1d6a165b33b02a09

src/backends/mod.rs
```diff
@@ -4,17 +4,23 @@ use ark_ff::{Field, Zero};
...
 
+use self::{kimchi::KimchiVesta, r1cs::R1csBls12_381};
+
 pub mod kimchi;
+pub mod r1cs;
+
+pub enum BackendKind {
+    KimchiVesta(KimchiVesta),
+    R1CS(R1csBls12_381),
+}
```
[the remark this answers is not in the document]
katat: See comment above ^

src/backends/r1cs/mod.rs
```diff
@@ -0,0 +1,557 @@
...
+        self.witness_vars.len() - self.public_inputs.len() - self.public_outputs.len()
+    }
+
+    // Helper function to convert &[u64] to BigInt
+    fn bigint_from_u64_slice(slice: &[u64]) -> BigInt {
+        let mut bytes = Vec::new();
+        for &num in slice.iter().rev() { // Reverse the slice to match the big-endian byte order
+            bytes.extend(&num.to_be_bytes());
+        }
+        BigInt::from_bytes_be(Sign::Plus, &bytes)
```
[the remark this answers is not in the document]
katat: actually there is function from the lib :P

src/backends/r1cs/mod.rs
katat: implemented from_vars :D

src/backends/r1cs/mod.rs
```diff
@@ -0,0 +1,489 @@
...
+            constraints: Vec::new(),
+            witness_vars: Vec::new(),
+            debug_info: Vec::new(),
+            public_inputs: Vec::new(),
+            public_outputs: Vec::new(),
+            finalized: false,
+        }
+    }
+
+    // todo: how can we create this r1cs backend with different associated field types, but still using the same backend implementation? using macro?
```
mimoo, line 122: I don't think you need to use macro, you could just have a `R1CS` struct generic on the prime, and then instantiate it in two different backends.

src/backends/r1cs/mod.rs
```diff
@@ -0,0 +1,489 @@
...
+    /// So that we can get the associated field type, instead of passing it as a parameter here.
+    /// There are two fields supported by the snarkjs for r1cs: bn128 and ark_bls12_381.
+    /// Currently we are using ark_bls12_381
+    fn prime(&self) -> BigInt {
+        BigInt::from_slice_native(Sign::Plus, Fr::characteristic())
+    }
+
+    /// Add an r1cs constraint that is 3 linear combinations.
+    /// This represents one constraint: a * b = c
+    pub fn add_constraint(
```
mimoo, line 132: does it need to be `pub`?

src/backends/r1cs/mod.rs
```diff
@@ -0,0 +1,489 @@
...
+    type Field = Fr;
+
+    type GeneratedWitness = GeneratedWitness;
+
+    fn witness_vars(&self, var: CellVar) -> &Value<Self> {
+        self.witness_vars.get(var.index).unwrap()
+    }
+
+    fn poseidon() -> crate::imports::FnHandle<Self> {
+        poseidon
```
mimoo, line 170: nit: personally I like `builtins::poseidon` better, but heh :D

src/backends/r1cs/mod.rs
```diff
@@ -0,0 +1,489 @@
...
+
+    /// Generate the witnesses
+    /// This process should check if the constraints are satisfied.
+    fn generate_witness(
+        &self,
+        witness_env: &mut crate::witness::WitnessEnv<Self::Field>,
+    ) -> crate::error::Result<Self::GeneratedWitness> {
+        if !self.finalized {
+            panic!("the circuit is not finalized yet!");
+        }
```
mimoo, line 257: btw you can just write `assert!(self.finalized, "the circuit is not finalized yet!");`

src/backends/r1cs/mod.rs
```diff
@@ -0,0 +1,489 @@
...
+
+        // generate witness through witness vars vector
+        let mut witness = self.witness_vars.iter().enumerate().map(|(index, val)| {
+            self.compute_val(witness_env, val, index)
+        }).collect::<crate::error::Result<Vec<Fr>>>()?;
```
mimoo, line 262: wait so, my understanding is that either the public_output values are `Value::PublicOutput(None)`, in which case it will crash. Or they are set, in which case they will be calculated correctly at this point.

If the circuit is finalized, they should be set, so we shouldn't have to compute them after no?

src/backends/r1cs/mod.rs
```diff
@@ -0,0 +1,489 @@
...
+
+        // The original vars of public outputs are not part of the constraints
+        // so we need to compute them separately
+        for var in &self.public_outputs {
+            let val = self.compute_var(witness_env, *var)?;
+            witness[var.index] = val;
```
mimoo, line 268: so I think this is unnecessary, I guess you can quickly check this by checking if the `public_output.no` leads to the same witness with this line or without (or if the witness verifies without this line)

src/backends/r1cs/mod.rs
mimoo: LGTM!

src/backends/r1cs/mod.rs
```diff
@@ -0,0 +1,489 @@
...
+        }
+
+        for constraint in &self.constraints {
+            // assert a * b = c
+            assert_eq!(
+                constraint.a.evaluate(&witness) * constraint.b.evaluate(&witness), 
+                constraint.c.evaluate(&witness)
+            );
+        }
```
mimoo, line 277: we should return a nice error to the user here, like we do in kimchi:

```rust
return Err(Error::new(
                                "runtime",
                                ErrorKind::InvalidWitness(row),
                                debug_info.span,
                            ));
```

src/backends/r1cs/mod.rs
```diff
@@ -0,0 +1,489 @@
...
+                if !self.public_outputs.iter().any(|&x| x.index == index) {
+                    panic!("there's a bug in the circuit_writer, some cellvar does not end up being a cellvar in the circuit!");
+                }
+            }
+        }
+
+        self.finalized = true;
+
+        Ok(())
+    }
```
mimoo, line 247: note to meself: re-read this one last time ^^

src/backends/r1cs/mod.rs
katat: The public output cell is replaced in the `finalize_circuit`:

https://github.com/zksecurity/noname/blob/77e44090d66f80177dccfe53af9ee81665ccdc22/src/backends/r1cs/mod.rs#L215-L221

So actually they can be calculated directly via this line of code, and the following lines are not necessary any more:
https://github.com/zksecurity/noname/blob/77e44090d66f80177dccfe53af9ee81665ccdc22/src/backends/r1cs/mod.rs#L266-L269

src/backends/r1cs/mod.rs
katat: Agreed

src/backends/r1cs/mod.rs
katat: removed all the unnecessary pubs

src/backends/r1cs/mod.rs
katat: maybe my reasoning in the below comment makes sense?

src/backends/r1cs/mod.rs
```diff
@@ -0,0 +1,489 @@
...
+        for constraint in &self.constraints {
+            for lc in constraint.as_array() {
+                for var in lc.terms.keys() {
+                    written_vars.insert(var.index);
+                }
+            }
+        }
+
+        // check if every cell vars end up being a cell var in the circuit or public output
+        for index in 0..self.witness_vars.len() {
```
mimoo, line 235: dangerous! always use iterators (`for witness_var in &self.witness_vars`)

mimoo (approved):
Everything looks great, besides the public output computation not being deferred. So I think this is approvable now :) feel free to merge and fix things later!

katat:
Updated to defer the output calculation and optimize the circuit finalization.

Issue #58, opened 2024-05-01
create a test to see if the verifier can insert anything in public output part of the public input and the proof still works


mimoo:
create a test to see if the verifier can insert anything in public output part of the public input and the proof still works

actually, not even proof, we should be able to test if a witness is correct without checking a proof

katat:
It is fine for kimchi.

In noname, it add a "half-constraint" when a new output is added.

https://github.com/zksecurity/noname/blob/b93d60453836523569cca7354541b0c449f0108c/src/backends/kimchi/mod.rs#L698-L703

Then this constraint will be cancelled by `- public` during verifying witness.

https://github.com/o1-labs/proof-systems/blob/b443013b9f6a2aa5e396b5d274b9899626fb850b/kimchi/src/circuits/polynomials/generic.rs#L300-L304

So if the verifier doesn't provide a correct public output value, that constraint won't be cancelled.

mimoo:
ohh right! We use the permutation in kimchi to wire these (since we copy the returned var up there in that part of the public input, they will be wired automatically)

for R1CS it won't work though, so it'd be good to have a test there also that shows that fails until we fix it :D

mimoo:
on double take, the check done by the verify function here is not enough, as it will always work (the public input cancels the public input). To make sure the permutation is the one that's doing the checking, we have to actually create a proof/verify 

(or have a better witness verify function, which we don't have at the moment ^^ but maybe it might be a good idea as well, but in the respect of time we should probably just create a proof and verify)

mimoo:
associated, this command indeed doesn't show any constraints OR wiring for the public output!

```
cargo run -- test --path examples/public_output.no --private-inputs '{"private_input": "1"}' --public-inputs '{"public_input": "1"}' --debug
```

Issue #55, opened 2024-04-26
use the permutation to constrain constraint_eq_var


mimoo:

mimoo:
I think the way we do the permutation right now is too naive to allow us to implement this optimization right away. We need to be able to keep track of cycles of variables, and to MERGE cycles of variables when we wire two variables that belong to two different cycles. We did this in snarky using union finds: https://github.com/o1-labs/proof-systems/blob/master/kimchi/src/snarky/constraint_system.rs#L528

I explain how this works here: https://github.com/o1-labs/proof-systems/blob/master/book/src/snarky/kimchi-backend.md#implementation-details

PR #59, opened 2024-05-02
Add test to check incorrect output for kimchi backend

katat:
https://github.com/zksecurity/noname/issues/58

The kimchi prover/ verifier already cover the public output constraints. Here we add a test to ensure this is the case.

### DIFF
--- a/examples/public_output.asm
+++ b/examples/public_output.asm
@@ -5,5 +5,8 @@ DoubleGeneric<1>
 DoubleGeneric<1,1,-1>
 DoubleGeneric<1,0,0,0,-2>
 DoubleGeneric<1,0,-1,0,6>
+DoubleGeneric<1,-1>
 (0,0) -> (2,1)
+(1,0) -> (5,0)
 (2,2) -> (3,0) -> (4,0)
+(4,2) -> (5,1)
--- a/src/backends/kimchi/mod.rs
+++ b/src/backends/kimchi/mod.rs
@@ -341,7 +341,6 @@ impl Backend for KimchiVesta {
                 let prev = self
                     .witness_vars
                     .insert(var_idx, Value::PublicOutput(Some(ret_var)));
-                // .insert(var_idx, Value::PublicOutput(Some(*ret_var)));
                 assert!(prev.is_some());
             }
         }
@@ -361,11 +360,10 @@ impl Backend for KimchiVesta {
 
         let mut witness = vec![];
         // compute each rows' vars, except for the deferred ones (public output)
-        let mut public_outputs_vars: Vec<(usize, CellVar)> = vec![];
+        let mut public_outputs_vars: HashMap<CellVar, Vec<(usize, usize)>> = HashMap::new();
 
-        for (row, (gate, row_of_vars, debug_info)) in
-            izip!(self.gates.iter(), &self.rows_of_vars, &self.debug_info).enumerate()
-        {
+        // calculate witness except for public outputs
+        for (row, row_of_vars) in self.rows_of_vars.iter().enumerate() {
             // create the witness row
             let mut witness_row = [Self::Field::zero(); NUM_REGISTERS];
 
@@ -373,7 +371,10 @@ impl Backend for KimchiVesta {
                 let val = if let Some(var) = var {
                     // if it's a public output, defer it's computation
                     if matches!(self.witness_vars()[&var.index], Value::PublicOutput(_)) {
-                        public_outputs_vars.push((row, *var));
+                        public_outputs_vars
+                            .entry(*var)
+                            .or_default()
+                            .push((row, col));
                         Self::Field::zero()
                     } else {
                         self.compute_var(witness_env, *var)?
@@ -384,7 +385,24 @@ impl Backend for KimchiVesta {
                 witness_row[col] = val;
             }
 
-            // check if the row makes sense
+            witness.push(witness_row);
+        }
+
+        // compute public output at last
+        let mut public_outputs = vec![];
+
+        for (var, rows_cols) in public_outputs_vars {
+            let val = self.compute_var(witness_env, var)?;
+            for (row, col) in rows_cols {
+                witness[row][col] = val;
+            }
+            public_outputs.push(val);
+        }
+
+        // sanity check the witness
+        for (row, (gate, witness_row, debug_info)) in
+            izip!(self.gates.iter(), &witness, &self.debug_info).enumerate()
+        {
             let is_not_public_input = row >= self.public_input_size;
             if is_not_public_input {
                 #[allow(clippy::single_match)]
@@ -414,18 +432,6 @@ impl Backend for KimchiVesta {
                     _ => (),
                 }
             }
-
-            //
-            witness.push(witness_row);
-        }
-
-        // compute public output at last
-        let mut public_outputs = vec![];
-
-        for (row, var) in public_outputs_vars {
-            let val = self.compute_var(witness_env, var)?;
-            witness[row][0] = val;
-            public_outputs.push(val);
         }
 
         // extract full public input (containing the public output)
--- a/src/backends/kimchi/prover.rs
+++ b/src/backends/kimchi/prover.rs
@@ -233,3 +233,59 @@ impl VerifierIndex {
         .wrap_err("kimchi: failed to verify the proof")
     }
 }
+
+mod testing {
+    use crate::{
+        backends::{
+            kimchi::{KimchiVesta, VestaField},
+            Backend,
+        },
+        constants::Span,
+        var::{ConstOrCell, Value, Var},
+        witness::WitnessEnv,
+    };
+
+    #[test]
+    fn test_public_output_constraint() -> miette::Result<()> {
+        // setup a simple circuit
+        let mut backend = KimchiVesta::new(false);
+
+        let span = Span::default();
+        let input_val = VestaField::from(1);
+        let input_var = backend.add_public_input(Value::Constant(input_val), span);
+
+        let output_var = backend.add_public_output(Value::PublicOutput(None), span);
+
+        // 1 + 1 = 2
+        let const_val = VestaField::from(1);
+        let res_val = input_val + const_val;
+        let res_var = backend.add_const(&input_var, &const_val, span);
+        let public_outputs = Some(Var::new_cvar(ConstOrCell::Cell(output_var), span));
+        let returned_cells = Some(Vec::from([res_var]));
+
+        backend
+            .finalize_circuit(public_outputs, returned_cells, vec![], span)
+            .unwrap();
+
+        // test with kimchi prover
+        let mut witness_env = WitnessEnv::default();
+        let generated_witness = backend.generate_witness(&mut witness_env).unwrap();
+        let witness = generated_witness.all_witness.to_kimchi_witness();
+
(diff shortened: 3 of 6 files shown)


src/backends/kimchi/prover.rs
```diff
@@ -233,3 +233,59 @@ impl VerifierIndex {
         .wrap_err("kimchi: failed to verify the proof")
     }
 }
+
+mod testing {
```
mimoo, line 237: `#[cfg(test)]`

src/backends/kimchi/prover.rs
```diff
@@ -233,3 +233,59 @@ impl VerifierIndex {
...
+        let witness = generated_witness.all_witness.to_kimchi_witness();
+
+        let (prover_index, _) = backend.compile_to_indexes().unwrap();
+
+        // correct output value
+        let mock_public_entries = [input_val, res_val];
+        prover_index.verify(&witness, &mock_public_entries).unwrap();
+
+        // incorrect output value
+        let mock_public_entries = [input_val, VestaField::from(1)];
```
mimoo, line 283: maybe `res_val + 1` would ensure that it fails :D

src/backends/kimchi/prover.rs
```diff
@@ -233,3 +233,59 @@ impl VerifierIndex {
...
+        let input_var = backend.add_public_input(Value::Constant(input_val), span);
+
+        let output_var = backend.add_public_output(Value::PublicOutput(None), span);
+
+        // 1 + 1 = 2
+        let const_val = VestaField::from(1);
+        let res_val = input_val + const_val;
+        let res_var = backend.add_const(&input_var, &const_val, span);
+        let public_outputs = Some(Var::new_cvar(ConstOrCell::Cell(output_var), span));
+        let returned_cells = Some(Vec::from([res_var]));
```
mimoo, line 265: I think it would have been nicer to compile something like this instead to make sure the end-to-end flow doesn't have a bug (and then it's easier to check any backend with it as well):

```rust
fn main(pub public_input: Field, private_input: Field) -> Field {
    let xx = private_input + public_input;
    assert_eq(xx, 2);
    let yy = xx + 6;
    return yy;
}
```

the second thing is that... the verify function doesn't check the permutation, so it's not going to detect the bug (if there is one) :D I believe we actually have to create a proof

mimoo: so something like that, roughly:

```rust
let code = r#"fn main(pub public_input: Field, private_input: Field) -> Field {
    let xx = private_input + public_input;
    assert_eq(xx, 2);
    let yy = xx + 6;
    return yy;
}";

let mut sources = Sources::new();
let mut tast = TypeChecker::new();
let this_module = None;
let _node_id = typecheck_next_file(
    &mut tast,
    this_module,
    &mut sources,
    "public_output.no".to_string(),
    code.clone(),
    0,
)
.unwrap();

let kimchi_vesta = KimchiVesta::new(false);
let compiled_circuit = compile(&sources, tast, kimchi_vesta)?;

let (prover_index, verifier_index) = compiled_circuit.compile_to_indexes().unwrap();

let generated_witness = generate_witness(
    &self.compiled_circuit,
    sources,
    public_inputs,
    private_inputs,
)?;

// TODO: modify the public output part of the witness

let witness = generated_witness.all_witness.to_kimchi_witness();

// verify the witness
self.index
    .verify(&witness, &generated_witness.full_public_inputs)
    .unwrap();

// create proof
let proof =
    ProverProof::create::<BaseSponge, ScalarSponge>(&GROUP_MAP, witness, &[], &self.index)
        .into_diagnostic()
        .wrap_err("kimchi: could not create a proof with the given inputs")?;

kimchi::verifier::verify::<Curve, BaseSponge, ScalarSponge, OpeningProof<Curve>>(
    &GROUP_MAP,
    &self.index,
    &proof,
    &full_public_inputs,
)
.into_diagnostic()
.wrap_err("kimchi: failed to verify the proof")    
```

katat:
I had a check, the verify function from the ProverIndex seems also including the permutation check? 

```rust
for col in 0..PERMUTS {
  let wire = gate.wires[col];

  if wire.col >= PERMUTS {
      return Err(GateError::Custom {
          row,
          err: format!("a wire can only be connected to the first {PERMUTS} columns"),
      });
  }

  if witness[col][row] != witness[wire.col][wire.row] {
      return Err(GateError::DisconnectedWires(
          Wire { col, row },
          Wire {
              col: wire.col,
              row: wire.row,
          },
      ));
  }
}
```

Once the `assert_eq` constraint for the output is added, this check ^ seems already covering without the need to generate proof. (The refactored test demonstrates this fact)

examples/public_output.asm
```diff
@@ -5,5 +5,8 @@ DoubleGeneric<1>
 DoubleGeneric<1,1,-1>
 DoubleGeneric<1,0,0,0,-2>
 DoubleGeneric<1,0,-1,0,6>
+DoubleGeneric<1,-1>
 (0,0) -> (2,1)
+(1,0) -> (5,0)
```
mimoo, line 10: shouldn't it be `(0,0) -> (5,0)`?

mimoo: oh no, that's the public input, `(1, 0)` is the public output :D

src/backends/kimchi/prover.rs
```diff
@@ -233,3 +233,113 @@ impl VerifierIndex {
...
+        // this is the gate contains the output var
+        assert_eq!(witness[0][1], VestaField::from(output_val));
+        // this is the assert_eq gate which result var and the output var
+        assert_eq!(witness[0][gate_count - 1], VestaField::from(output_val));
+
+        // should pass the sanity check
+        prover_index
+            .index
+            .verify(&witness, &full_public_inputs)
+            .unwrap();
```
mimoo, line 323: can you also check that this returns an error as well if we modify the pub output then? (if what you say is correct about the permutation I guess this should work as well)

src/backends/kimchi/prover.rs
```diff
@@ -233,3 +233,113 @@ impl VerifierIndex {
...
+
+        assert_eq!(
+            generated_witness.full_public_inputs,
+            vec![
+                VestaField::from(public_input_val),
+                VestaField::from(output_val)
+            ]
+        );
+
+        // modify the output value
```
mimoo, line 308: you don't modify anything here

src/circuit_writer/mod.rs
```diff
@@ -223,6 +223,17 @@ impl<B: Backend> CircuitWriter<B> {
...
         let public_output = circuit_writer.public_output.clone();
 
+        // constraint public outputs to the result of the circuit
+        if let Some(public_output) = &public_output {
+            let cvars = &public_output.cvars;
+
+            for (pub_var, ret_var) in cvars.iter().zip(&returned_cells.clone().unwrap()) {
+                circuit_writer
+                    .backend
+                    .assert_eq_var(pub_var.cvar().unwrap(), ret_var, main_span);
```
mimoo, line 233: beautiful 🤩

src/backends/kimchi/mod.rs
```diff
@@ -384,7 +385,24 @@ impl Backend for KimchiVesta {
...
+            for (row, col) in rows_cols {
+                witness[row][col] = val;
+            }
+            public_outputs.push(val);
+        }
+
+        // sanity check the witness
+        for (row, (gate, witness_row, debug_info)) in
+            izip!(self.gates.iter(), &witness, &self.debug_info).enumerate()
+        {
```
mimoo, line 405: btw what is this change related to in this file?

mimoo (commented):
Ah I see! I guess it would be nice to also check that the witness verification function returns an error when we modify the public output then :o 

best practice is to include a screenshot of the test failing _before_ you fix the test by adding the constraints, can you do that :D?

also I didn't review some code that seems unrelated to this PR, so wondering if I need to?

src/backends/kimchi/prover.rs
katat: it does this check here:
https://github.com/zksecurity/noname/blob/444ca6a2c8f19318f224ed64e93b315c96b938c4/src/backends/kimchi/prover.rs#L325-L341

perhaps we can also make it assert it returns `DisconnectedWires ` error (see the above code snippet) if the output is modified.

src/backends/kimchi/mod.rs
katat: The `generate_witness` function was assuming there is only a `PublicOutput` var in the circuit. But after constraining the output to the result of circuit, this public output cell will appear twice at least in the circuit:
 - One time as the public output constraint (as part of the public entry), which will be cancelled by the verifier
 - Second time as the part of the constraint with the result of the circuit

So here we need to have a hash map to record where are the places this output cell appears in circuit. With this location info, the output value calculated at the last via `compute_var` can be put to the correct `<row, col>`.

katat:
> best practice is to include a screenshot of the test failing before you fix the test by adding the constraints, can you do that :D?

I rebased the commits to showcase the failed test at the commit https://github.com/zksecurity/noname/pull/59/commits/8ff921e966f743b7464199d3252d6daff0d88ff7

screenshot: 
Screenshot 2024-05-06 at 10 22 34

Then it is fixed by the subsequent commit, which details the test on different types of verification errors.

> also I didn't review some code that seems unrelated to this PR, so wondering if I need to?

I think they are related to the fix. Please refer to my comments above :D

src/backends/kimchi/mod.rs
```diff
@@ -361,19 +360,21 @@ impl Backend for KimchiVesta {
...
-            izip!(self.gates.iter(), &self.rows_of_vars, &self.debug_info).enumerate()
-        {
+        // calculate witness except for public outputs
+        for (row, row_of_vars) in self.rows_of_vars.iter().enumerate() {
             // create the witness row
             let mut witness_row = [Self::Field::zero(); NUM_REGISTERS];
 
             for (col, var) in row_of_vars.iter().enumerate() {
                 let val = if let Some(var) = var {
                     // if it's a public output, defer it's computation
```
mimoo, line 372: perhaps here we should add a comment "to avoid deep recursion"

src/backends/kimchi/mod.rs
mimoo: oh right... I was thinking about it like we're using the permutation for `assert_eq`, but we still don't :D (cf https://github.com/zksecurity/noname/issues/55)

src/backends/kimchi/mod.rs
```diff
@@ -361,19 +360,21 @@ impl Backend for KimchiVesta {
 
         let mut witness = vec![];
         // compute each rows' vars, except for the deferred ones (public output)
-        let mut public_outputs_vars: Vec<(usize, CellVar)> = vec![];
+        let mut public_outputs_vars: HashMap<CellVar, Vec<(usize, usize)>> = HashMap::new();
```
mimoo, line 363: can you use the `Cell` struct instead of `(usize, usize)` here?

src/backends/kimchi/mod.rs
```diff
@@ -361,19 +360,21 @@ impl Backend for KimchiVesta {
 
         let mut witness = vec![];
```
mimoo, line 361: unrelated: this really needs a `Vec::with_capacity(self.rows_of_vars.len())`

mimoo (approved):
awesome, makes sense to me now :) thanks for taking the time to explain!

PR #62, opened 2024-05-07
Support R1CS asm

katat:
This PR is to support asm (a debug feature) for R1CS backends.

Currently here is a draft example for equals.no: 
```
v_0 + v_3 = 0
v_1 + v_3 = v_4
v_2 + v_5 = 0
v_5 + 1 = v_6
(v_2) * (v_4) = v_7
v_7 = 0
(v_8) * (v_4) = v_9
v_9 = v_6
v_2 = 1
v_10 = 3
v_0 + v_12 = 0
v_12 + v_10 = v_13
v_11 + v_14 = 0
v_14 + 1 = v_15
(v_11) * (v_13) = v_16
v_16 = 0
(v_17) * (v_13) = v_18
v_18 = v_15
v_11 = 1
```

`v_*` are the witness variables.

Each line represents a constraint (a*b = c) that contains the linear combinations of terms made up of factor and witness variables. If the value is 1 for either a or b, the linear combination will be ignored for reducing the unnecessary info.

- [x] review the format
- [x] wire up with span of debug info
- [ ] move generate_asm out of backend trait


mimoo:
Quick comment: this format is not bijective. There are several r1cs constraints system that correspond to this one (for example, each line that doesn't involve a multiplication could have the constraint be encoded in A, B or C)

katat:
Updated the format. Here is the example for public_output.no

```
v_3 == (v_0 + v_1) * (1)
2 == (v_3) * (1)
v_4 == (v_3 + 6) * (1)
v_4 == (v_2) * (1)

```

katat:
> I'm also not sure I understand it from your example, is it C == A * B? If so, why not place A * B first and then C? (so A * B == C) which is the "convention" when I see R1CS

It was intended to align with the noname code in debug mode :D

Yeah, let's use A * B == C instead.

mimoo:
also let's write `1` instead of `v_0`? (if the intention is to have v_0 be 1?)

src/tests/examples.rs
```diff
@@ -108,6 +110,23 @@ fn test_file(
             compiled_circuit
                 .generate_witness(public_inputs.clone(), private_inputs.clone())
                 .unwrap();
+
+            // check the ASM
+            if compiled_circuit.circuit.backend.len() < 100 {
```
mimoo, line 115: btw not a fan of this `len()` function. IMO it should have a better name (like `num_constraints`).

katat:
> also let's write `1` instead of `v_0`? (if the intention is to have v_0 be 1?)

do you mean substituting the witness variable with its value? but will this lost the trace of the variables?

src/lib.rs
```diff
@@ -55,11 +56,99 @@ pub mod helpers {
...
+qrst
+uvwx
+yz
+";
+            assert_eq!(
+                find_exact_line(SRC, crate::constants::Span::new(0, 5, 6)),
+                (2, 5, "efgh\nijkl")
+            );
+        }
+    }
```
mimoo, line 153: btw not a fan of having logic in `lib.rs` in general. I think if we have a problem having too much helper functions and we don't know where to have them, then we should create a `utils` module and move these kind of things there

src/cli/cmd_build_and_check.rs
```diff
@@ -234,6 +240,7 @@ pub struct CmdTest {
     #[clap(short, long)]
     debug: bool,
 
+    // todo: this is kimchi specific and it doesn't work atm. we might just remove this option at this stage?
```
mimoo, line 243: I'm fine if you remove it and disable the double generic gate by default (since kimchi is deprioritized)

src/cli/cmd_build_and_check.rs
```diff
@@ -265,34 +274,59 @@ pub fn cmd_test(args: CmdTest) -> miette::Result<()> {
...
-    let (prover_index, verifier_index) = compiled_circuit.compile_to_indexes()?;
-    println!("successfully compiled");
-
-    // print ASM
-    let asm = prover_index.asm(&sources, args.debug);
-    println!("{asm}");
+    match backend.as_str() {
+        "kimchi-vesta" => {
+            let mut tast = TypeChecker::new();
+            let _node_id = typecheck_next_file(
```
mimoo, line 281: I'm always confused when I look at this code, where we have to init Sources, then init typechecker, then call typecheck_next_file, and then I'm wondering which one do we call if we have several files :D

I'm not sure what would be a good alternative, but we should probably create an issue to brainstorm that. First, I think we should have convenience functions in the compiler for any usecase (compile a single file, compile a single file from source, compile multiple files from source, etc.) 

second, maybe we could have some sort of builder pattern to make the flow much clearer. Anyway it's not super high priority but my brain always thinks of cleaning up ^^

examples/fixture/asm/r1cs/poseidon.asm
```diff
@@ -0,0 +1,185 @@
+@ noname.0.7.0
```
mimoo, line 1: this file is wrong, not sure why it's displaying kimchi stuff. Is there a clean way to ignore this one in R1CS tests atm?

mimoo (approved):
LGTM!

examples/fixture/asm/r1cs/poseidon.asm
katat: Ah, the poseidon builtin hasn't been implemented for r1cs and this file was copied from the kimchi asm.

src/cli/cmd_build_and_check.rs
katat: That would be better: https://github.com/zksecurity/noname/issues/67

PR #63, opened 2024-05-10
Refactor R1CS struct to be generic on field

katat:


src/backends/mod.rs
```diff
@@ -7,25 +7,50 @@ use crate::{
...
+pub trait BackendField:
+    Field + FromStr + TryFrom<BigUint> + TryInto<BigUint> + Into<BigUint> + PrettyField
+{
+}
+
 pub enum BackendKind {
     KimchiVesta(KimchiVesta),
-    R1CS(R1csBls12_381),
+    R1csBls12_381(R1CS<R1csBls12381Field>),
+    R1csBn128(R1CS<R1csBn128Field>),
```
mimoo, line 33: should we call it bn254 everywhere? what's the most popular name :D?

katat: snarkjs calls it bn128 :)

katat: maybe better with bn254 as the google results seem to prefer this one

src/backends/r1cs/mod.rs
```diff
@@ -287,7 +312,13 @@ impl Backend for R1csBls12_381 {
             }
         }
 
-        Ok(GeneratedWitness { witness })
+        let outputs = self
+            .public_outputs
+            .iter()
+            .map(|var| witness[var.index])
+            .collect();
```
mimoo, line 319: Ah yeah! Indeed that is practical

src/backends/r1cs/mod.rs
```diff
@@ -121,17 +142,14 @@ impl R1csBls12_381 {
...
     }
 
-    // todo: how can we create this r1cs backend with different associated field types, but still using the same backend implementation? using macro?
-    /// So that we can get the associated field type, instead of passing it as a parameter here.
-    /// There are two fields supported by the snarkjs for r1cs: bn128 and ark_bls12_381.
-    /// Currently we are using ark_bls12_381
+    /// Returns the prime for snarkjs based on the curve field.
     fn prime(&self) -> BigInt {
-        BigInt::from_slice_native(Sign::Plus, Fr::characteristic())
+        BigInt::from_slice_native(Sign::Plus, F::characteristic())
```
mimoo, line 147: I don't think [characteristic](https://en.wikipedia.org/wiki/Characteristic_(algebra)) is the right function here btw. 

I think the problem of this function is that it assumes a `PrimeField` (which has a method to give you the prime) and not a `Field`

so my recommendation would be to expect a `F: BackendField + PrimeField` here and do `F::Params::MODULUS`

mimoo (approved):
btw we should update the README at some point ^^

src/backends/r1cs/mod.rs
```diff
@@ -82,25 +91,34 @@ impl LinearCombination {
...
-pub struct R1csBls12_381 {
+pub struct R1CS<F>
+where
+    F: BackendField,
+{
     /// Constraints in the r1cs.
-    constraints: Vec<Constraint>,
-    witness_vars: Vec<Value<R1csBls12_381>>,
+    constraints: Vec<Constraint<F>>,
+    witness_vars: Vec<Value<Self>>,
```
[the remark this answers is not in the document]
katat: maybe at the point when we are implementing the stdlib, so we can review the better way to handle `Hint`.

src/cli/cmd_build_and_check.rs
```diff
@@ -338,31 +341,68 @@ pub fn cmd_run(args: CmdRun) -> miette::Result<()> {
...
+                prover_index.prove(&sources, public_inputs, private_inputs, false)?;
+            println!("proof created");
 
-            snarkjs_exporter.gen_wtns_file(
-                &curr_dir.join("output.wtns").into_string(),
-                generated_witness,
-            );
+            // verify proof
+            verifier_index.verify(full_public_inputs, proof)?;
+            println!("proof verified");
```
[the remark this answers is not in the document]
katat: That makes sense as this run command is expected to generate circuit and witness files for r1cs, but it currently does inline proving and verifying for kimchi without generating anything. We may revisit this when it is needed to serialize the kimchi circuit and witness.

I updated it to keep `unimplemented!` for the kimchi backend for this command.

PR #64, opened 2024-05-10
Optimize constraints for r1cs

katat:
The whole idea is to have `CellVar` to accumulate the binary operations until it is necessary to enforce a constraint.

In order to do so, here are the todos:
- [x] refactor the CellVar to be an associated type for backend trait
- [x] Use CellVar to hold the internal var and linear combination 
- [ ] Replace the backend constraint interfaces with CellVar


katat:
Here is a draft to have a quick review to see if this is right direction.

## R1csCellVar
https://github.com/zksecurity/noname/blob/289b13c3ed2b23552d5e09560d092594e3bfc8d9/src/backends/r1cs/mod.rs#L35-L38
The current implementation allows the `CellVar` to be an enum as either a `WitnessVar` or a `LinearCombination`

## Recursive linear combination
https://github.com/zksecurity/noname/blob/289b13c3ed2b23552d5e09560d092594e3bfc8d9/src/backends/r1cs/mod.rs#L144-L150

The variables inside a `LinearCombination` is of `R1csCellVar` type. This means a linear combination can either recursively contain another linear combinations or witness vars: 

## ExpandedLinearCombination
https://github.com/zksecurity/noname/blob/289b13c3ed2b23552d5e09560d092594e3bfc8d9/src/backends/r1cs/mod.rs#L217-L251

When adding the linear combinations as a constraint, they will be converted to `ExpandedLinearCombination`, which is a `expanded` version of the symbolic linear combination. I think the benefit of this approach is to simply the logic during accumulating the linear combinations, and handle them all in one place (within the `expand` function).

For example, the following logic doesn't have to deal with specific cell var type, as they are wrapped as linear combinations which can be expanded at the end.
https://github.com/zksecurity/noname/blob/4f6239b8c1146b7de49bae02e67f5fee8c0fc7e3/src/backends/r1cs/mod.rs#L75-L108

## Linear combination calculation
This optimized R1cs constraint version uses the `CellVar::LinearCombination` for calculating the value for linear combination, instead of relying on the `Value::LinearCombination`.

https://github.com/zksecurity/noname/blob/4f6239b8c1146b7de49bae02e67f5fee8c0fc7e3/src/backends/r1cs/mod.rs#L646-L664

## Debug info
After the constraint is optimized, it might be helpful to keep the trace of how it comes up with the constraint and reduce the repeated info.

For example:

Refine the following(this is what it current shows):
```
v_4 == (v_3) * (-1 * v_1 + v_2)
v_6 == (v_5) * (-1 * v_1 + v_2)
```

To:
```
lc = -1 * v_1 + v_2
v_4 == (v_3) * (lc_1)
v_6 == (v_5) * (lc_1)
```

So while it can still trace where the `lc` is from, it avoids repeated string behind `lc`.

katat:
The LinearCombination is reverted to non-recursive version, that is it only contains the witness vars as the part of terms instead of allowing recursive structure of LinearCombination.

The debug spans in the backend arithmetic methods are left unused until we decide if it is necessary to refactor the way we store and track the LinearCombination. The current implementation of LinearCombination doesn't track itself (it just updates its own terms after `add` or `mul` operations, then return the new LinearCombination). We could just display these linear combinations for now, although displayed asm info can be redundant (as I tried to explain above ^) for some lines of code since it doesn't track where the intermediate linear combinations (the intermediate states of a LinearCombination before being reduced by a constraint).

One way to track the LinearCombination, we might need r1cs backend to have a vector for the intermediate linear combinations similar to the vector of witness vars.

The other issue is that the refactor of R1csCellVar lead to the use of clone() whenever the R1csCellVar is presented. That is due to the fact the R1csCellVar can contains linear combination that holds hash map for terms. This could be avoid if the backend track the linear combination vector ^, then the R1csCellVar only needs to have a couple enum primitive types while the data can be retrieved from the vector managed by the R1CS backend.

src/backends/r1cs/mod.rs
```diff
@@ -379,13 +390,24 @@ where
...
+            R1csCellVar::LinearCombination(lc) => {
+                let mut val = lc.constant;
+
+                for (wv, factor) in lc.terms {
+                    let wv_val = self.witness_vars.get(wv.index).unwrap();
+                    let calc = self.compute_val(env, wv_val, wv.index)? * factor;
+                    val += calc;
+                }
+                
+                Ok(val)
```
mimoo, line 425: mmm, I'm wondering if you should panic if it's a linear combination here. (Or if you refactor the code as I explained, then we should always expect a single term (that's not the first variable) and a 0 constant

src/backends/r1cs/mod.rs
```diff
@@ -507,229 +529,59 @@ where
...
-        // c = cst
-
-        let a = LinearCombination::from_vars(vec![*x]);
-        let b = LinearCombination::one();
-        let c = LinearCombination::from_const(cst);
+    fn assert_eq_const(&mut self, x: &R1csCellVar<F>, cst: F, span: crate::constants::Span) {
+        let c = R1csCellVar::LinearCombination(LinearCombination::from_const(cst));
 
-        self.add_constraint("eq constraint: x = cst", Constraint { a, b, c }, span);
+        x.eq(self, &c)
```
mimoo, line 595: assert_eq

mimoo (approved):
approved, but two important comments:

* where is the first cell set as 1
* I think the naming and the structs can be improved

not sure I see how debugging is affected, could be nice to dump an example to see what it looks like :)

as for optimization by tracking linear combination in a map inside the circuitwriter, can you write an issue for that pointing to the relevant code in arkworks or elsewhere where they do that? I think it's low priority for now, but it could be interesting to have someone implement the change

src/backends/r1cs/mod.rs
```diff
@@ -23,12 +23,18 @@ impl BackendField for R1csBls12381Field {}
 impl BackendField for R1csBn254Field {}
 
 #[derive(Default, Clone, Copy, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
-pub struct R1csCellVar {
+pub struct WitnessVar {
```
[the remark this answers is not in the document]
katat: Refactor following this idea.

I keep the LinearCombination, trying to make the structure behind R1csVar bit more explicit via the naming.

src/backends/r1cs/mod.rs
```diff
@@ -105,6 +105,37 @@ where
...
+        for (var, factor) in other.terms.iter() {
+            let exists = terms.get(var);
+            match exists {
+                Some(f) => {
+                    let new_factor = *f + factor;
+                    terms.insert(*var, new_factor);
+                },
+                None => {
+                    terms.insert(*var, *factor);
+                }
```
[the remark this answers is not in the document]
katat: this is so elegant!

src/backends/r1cs/mod.rs
```diff
@@ -36,9 +36,61 @@ pub enum R1csCellVar<F: BackendField> {
     LinearCombination(LinearCombination<F>)
 }
 
-impl R1csCellVar {
-    fn new(index: usize, span: Span) -> Self {
-        Self { index, span }
+impl<F: BackendField> CellVar for R1csCellVar<F> {}
```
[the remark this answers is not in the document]
katat: Refactored!

src/backends/r1cs/mod.rs
```diff
@@ -275,6 +275,14 @@ where
     fn private_input_number(&self) -> usize {
         self.witness_vars.len() - self.public_inputs.len() - self.public_outputs.len()
     }
+
+    fn enforce_constraint(&mut self, a_cvar: &R1csCellVar<F>, b_cvar: &R1csCellVar<F>, c_cvar: &R1csCellVar<F>) {
+        let a = a_cvar.clone().into_linear_combination();
+        let b = b_cvar.clone().into_linear_combination();
+        let c = c_cvar.clone().into_linear_combination();
+
+        self.add_constraint("enforce constraint", Constraint { a, b, c }, Span::default());
```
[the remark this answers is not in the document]
katat: It should have span. Updated

src/backends/r1cs/mod.rs
katat: Refactored

src/backends/r1cs/mod.rs
```diff
@@ -319,7 +330,7 @@ where
         label: Option<&'static str>,
         value: F,
         span: crate::constants::Span,
-    ) -> R1csCellVar {
+    ) -> R1csCellVar<F> {
```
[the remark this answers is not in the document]
katat: https://github.com/zksecurity/noname/issues/74

katat:
> not sure I see how debugging is affected, could be nice to dump an example to see what it looks like :)

The intermediate calculations from the linear combinations are skipped. Take https://github.com/zksecurity/noname/blob/0c887067ba9387e76fa5ef28168c067c13cbf79b/examples/arithmetic.no#L2 for example, it only displays the span for the constraints.

Screenshot 2024-05-28 at 14 57 48


I have updated the [linear combination struct](https://github.com/zksecurity/noname/pull/64/commits/2ea1773a71cd7a478d7ff9a308f5fb2570d4abd5#diff-dfeb8bc5dde0542bebd85f4dd3a0b86e4c43326cfbc4f9aed589f40f28c223d7R52) to also include the span. So we would just update the debug formatting to also display the intermediate linear combinations along with source code spans: https://github.com/zksecurity/noname/issues/75

src/backends/r1cs/mod.rs
```diff
@@ -28,72 +28,15 @@ pub struct CellVar {
...
-    /// Enforces a constraint for the equality of two CellVars.
-    /// It needs to constraint: self * 1 = other.
-    fn assert_eq(&self, cs: &mut R1CS<F>, other: &Self, span: Span) {
-        let one_cvar = R1csCellVar::LinearCombination(LinearCombination::one());
-
-        cs.enforce_constraint(self, &one_cvar, other, span)
+impl CellVar {
+    /// Convert to linear combination.
+    fn to_linear_combination<F: BackendField>(self) -> LinearCombination<F> {
+        LinearCombination::from_vars(vec![self], self.span)
```
mimoo, line 34: at this point it might be better to just have

```rust
impl<F> From<CellVar> for LinearCombination<F> {
  // ...
}
```

src/backends/r1cs/mod.rs
```diff
@@ -126,40 +85,44 @@ where
...
         }
     }
 
     /// Create a linear combination to represent constant zero.
-    fn zero() -> Self {
+    fn zero(span: Span) -> Self {
         LinearCombination {
             terms: HashMap::new(),
             constant: F::zero(),
+            span,
```
mimoo, line 101: it's always hard for me to think of "what should have a span"? I think the real answer is "this needs a span because otherwise the error I'm seeing right now is not great"

src/backends/r1cs/mod.rs
```diff
@@ -357,16 +337,16 @@ where
...
+        main_span: Span,
     ) -> crate::error::Result<()> {
         // store the return value in the public input that was created for that
         if let Some(public_output) = public_output {
             let cvars = &public_output.cvars;
 
             for (pub_var, ret_var) in cvars.clone().iter().zip(returned_cells.unwrap()) {
                 // replace the computation of the public output vars with the actual variables being returned here
-                let var_idx = pub_var.cvar().unwrap().to_witness_var().index;
+                let var_idx = pub_var.cvar().unwrap().to_cell_var().index;
```
mimoo, line 345: I don't see why this would always be true, for example if you write something like this in noname:

```rust
fn main(pub xx: Field, yy: Field) -> Field {
   return xx + yy;
}
```

then the returned cell will be a linear combination of two vars, and the call to `to_cell_var()` will panic.

actually, let's merge this as is, and let's write a test to see if we can trigger the bug, wdyt?

mimoo: if I'm right, then the fix will probably be to reduce the linear combination to a single var before this for loop

mimoo (approved):
Looks great to me! The only thing I don't see answered is the first public input being set to 1, what happened to this?

src/backends/r1cs/mod.rs
katat: I think the span here would be helpful to debug the intermediate linear combinations (non constraints).

src/backends/r1cs/mod.rs
katat: I thought there will be always a single var for a public output, while the `returned_cells` holds the value (xx + yy). 

I will do a check

katat:
That will be tackled in another PR: https://github.com/zksecurity/noname/issues/70

Issue #56, opened 2024-04-30
provide a sub function in the backend


mimoo:
instead of doing neg and add

Issue #38, opened 2024-04-16
Can we remove `Value::Hint`?


mimoo:
`Value::Hint` allows us to write more complex functions when computing out-of-circuit in the compiler. But it is annoying because we can't easily serialize these functions.

For example, we might want to write a js (or rust-wasm) interpret that generates the witness based on a list of `Value`s. So that one can use their browser to generate witnesses, for example. The usage of closures in `Value::Hint` makes that hard.

I think what we could do is have another enum inside of `Hint` which list all of the possible functions that we're implementing as more complex functions. Or we could also inline all of the possible hint functions within the `Value` enum.

mimoo:
(this would allow us to remove the type parameter on the backend)

PR #69, opened 2024-05-16
Refactor using backend associated CellVar trait

katat:


src/witness.rs
```diff
@@ -40,7 +40,7 @@ pub struct CompiledCircuit<B: Backend> {
     pub circuit: CircuitWriter<B>,
 }
 
-impl<B: Backend> CompiledCircuit<B> {
+impl<B: Backend + 'static> CompiledCircuit<B> {
```
mimoo, line 43: why 'static here? I'm always wondering what that implies exactly

src/var.rs
```diff
@@ -21,17 +21,17 @@ use crate::{
...
+// pub struct CellVar {
+//     pub index: usize,
+//     pub span: Span,
+// }
+
+// impl CellVar {
+//     pub fn new(index: usize, span: Span) -> Self {
+//         Self { index, span }
+//     }
+// }
```
mimoo, line 34: delete

src/var.rs
```diff
@@ -126,38 +127,39 @@ impl<F: Field> ConstOrCell<F> {
 
     pub fn idx(&self) -> Option<usize> {
```
mimoo, line 128: perhaps we should try to delete this

src/circuit_writer/writer.rs
```diff
@@ -113,12 +116,8 @@ impl Ord for AnnotatedCell {
 // Circuit Writer (also used by witness generation)
 //
 
-impl<B: Backend> CircuitWriter<B> {
-    fn compile_stmt(
-        &mut self,
-        fn_env: &mut FnEnv<B::Field>,
-        stmt: &Stmt,
-    ) -> Result<Option<VarOrRef<B>>> {
+impl<B: Backend + 'static> CircuitWriter<B> {
```
mimoo, line 119: why the addition of 'static here

src/circuit_writer/mod.rs
```diff
@@ -132,7 +127,7 @@ impl<B: Backend> CircuitWriter<B> {
     }
 }
 
-impl<B: Backend> CircuitWriter<B> {
+impl<B: Backend + 'static> CircuitWriter<B> {
```
mimoo, line 130: same here

src/circuit_writer/fn_env.rs
```diff
@@ -2,16 +2,16 @@ use std::collections::HashMap;
 
 use ark_ff::Field;
 
-use crate::{parser::types::TyKind, var::Var};
+use crate::{backends::Backend, parser::types::TyKind, var::Var};
 
 /// Information about a variable.
 #[derive(Debug, Clone)]
-pub struct VarInfo<F>
+pub struct VarInfo<B>
```
mimoo, line 9: maybe `VarInfo<F, CellVar>` is better, since we do that with `Var` already

mimoo (commented):
left you some comments. The big question I have is: can we make `CellVar` completley opaque to the outside? Intuitively it should work, I don't get why we need to expose any `usize` internals :o

katat:
Here are the places requiring the index of cell var. We can think of a way to make them opaque.

https://github.com/zksecurity/noname/blob/8196e389535fd60889432561da2729d34a472cc6/src/var.rs#L128-L133

https://github.com/zksecurity/noname/blob/8196e389535fd60889432561da2729d34a472cc6/src/backends/mod.rs#L113-L119

https://github.com/zksecurity/noname/blob/8196e389535fd60889432561da2729d34a472cc6/src/circuit_writer/writer.rs#L739

katat:
If we can [get rid of the Value::Hint](https://github.com/zksecurity/noname/issues/38), adding `static might not be necessary, and the `compute_var` can be completely lived in the backend.

The challenge I think is how we can replace the following without a `Value::Hint`:
https://github.com/zksecurity/noname/blob/8196e389535fd60889432561da2729d34a472cc6/src/constraints/field.rs#L196-L207

https://github.com/zksecurity/noname/blob/8196e389535fd60889432561da2729d34a472cc6/src/constraints/field.rs#L335-L349

https://github.com/zksecurity/noname/blob/8196e389535fd60889432561da2729d34a472cc6/src/backends/kimchi/builtin.rs#L99-L117

mimoo:
for the Hint::Value I propose that we either find linear combinations to express these OR add a new Value::Poseidon() => which contains the logic

src/backends/kimchi/mod.rs
```diff
@@ -715,3 +737,12 @@ impl Backend for KimchiVesta {
...
 }
+
+#[derive(Clone, Default, Debug, Serialize, Deserialize)]
+pub(crate) struct PendingGate {
+    pub label: &'static str,
+    #[serde(skip)]
+    pub coeffs: Vec<VestaField>,
+    pub vars: Vec<Option<KimchiCellVar>>,
+    pub span: Span,
+}
```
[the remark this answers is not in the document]
katat: This was intended to shift the kimchi related code to kimchi backend.

src/backends/r1cs/mod.rs
```diff
@@ -616,6 +624,14 @@ where
         var
     }
 
+    /// Adds the private input cell vars.
+    fn add_private_input(&mut self, val: Value<Self>, span: Span) -> Self::CellVar {
+        let var = self.new_internal_var(val, span);
+        self.private_input_indices.push(var.index);
+
+        var
+    }
```
mimoo, line 638: great!

src/backends/r1cs/mod.rs
```diff
@@ -275,10 +275,18 @@ where
...
+                    let err = Error::new(
+                        "constraint-finalization",
+                        ErrorKind::PrivateInputNotUsed,
+                        main_span,
+                    );
+                    return Err(err);
+                } else {
+                    panic!("there's a bug in the circuit_writer, some cellvar does not end up being a cellvar in the circuit!");
+                }
+            }
```
mimoo, line 289: good catch!

src/backends/mod.rs
```diff
@@ -119,25 +119,20 @@ pub trait Backend: Clone {
...
+    fn compute_var(
         &self,
-        env: &mut WitnessEnv<Self::Field>,
-        val: &Value<Self>,
-        var_index: usize,
+        env: &mut WitnessEnv<Self::Field, Self::CellVar>,
+        var: Self::CellVar,
     ) -> Result<Self::Field> {
-        if let Some(res) = env.cached_values.get(&var_index) {
+        if let Some(res) = env.cached_values.get(&var) {
```
mimoo, line 127: I'm a bit confused as how this cache would work now, mmmm

src/backends/r1cs/mod.rs
```diff
@@ -312,7 +312,12 @@ where
                     // Defer calculation for output vars.
                     // The reasoning behind this is to avoid deep recursion potentially triggered by the public output var at the beginning.
                     Value::PublicOutput(_) => Ok(F::zero()),
-                    _ => self.compute_val(witness_env, val, index),
+                    _ => {
+                        // wrap as CellVar
+                        // todo: we might have a cache for CellVar somewhere to retrieve the span by var index
+                        let var = R1csCellVar::new(index, Span::default());
+                        self.compute_var(witness_env, var)
```
mimoo, line 319: this feels very hacky to me

mimoo: my intuition is that the cache is not going to work properly because of the vars not necessarily matching (because of a different span, for example)

I think we need to use a usize for the cache, as such I think we need to move `compute_var` to the backend, does that make sense?

src/constraints/field.rs
```diff
@@ -279,81 +267,10 @@ pub fn if_else_inner<B: Backend>(
         ConstOrCell::Cell(cond) => *cond,
```
mimoo, line 267: why return a value here?

src/constraints/field.rs
```diff
@@ -279,81 +267,10 @@ pub fn if_else_inner<B: Backend>(
...
-            let cond_mul_then_m_else = mul(compiler, cond, &then_m_else, span)[0];
-            let res_to_check = add(compiler, &cond_mul_then_m_else, else_, span)[0];
-            compiler
-                .backend
-                .assert_eq_var(&res, res_to_check.cvar().unwrap(), span);
-
-            Var::new_var(res, span)
-        }
-    }
+    // determine the res via arithemtic
```
mimoo, line 270: typo

src/constraints/field.rs
```diff
@@ -190,44 +190,32 @@ fn equal_cells<B: Backend>(
...
-                    } else {
-                        Ok(B::Field::zero())
-                    }
-                })),
-                span,
-            );
-
             // 1. diff = x2 - x1
             let neg_x1 = compiler.backend.neg(&x1, span);
             let diff = compiler.backend.add(&x2, &neg_x1, span);
```
mimoo, line 195: I think we could have

```rust
trait Backend {
  fn sub(a: Self::CellVar, b: Self::CellVar, span) -> Self::CellVar { Self::add(a, Self::neg(b, span), span) }
}
```

src/constraints/field.rs
```diff
@@ -190,44 +190,32 @@ fn equal_cells<B: Backend>(
...
-            let res_mul_diff = compiler.backend.mul(&res, &diff, span);
-
-            compiler.backend.assert_eq_const(&res_mul_diff, zero, span);
-
-            // 4. diff_inv * diff = one_minus_res
             let diff_inv = compiler
                 .backend
                 .new_internal_var(Value::Inverse(diff), span);
 
+            // 2. diff_inv * diff = one_minus_res
```
mimoo, line 200: wrong comment I think

mimoo (commented):
LGTM except the cache, which I don't think will work

src/backends/r1cs/mod.rs
katat: The problem is the `compute_var` contains the shared code to calculate the values across different backends. Moving the `compute_var` might mean making this part duplicated in different backends.

katat: We might still need to keep the compute_val functions, and have the backends to implement the compute_var. 

So this allows the backends to decide how to load the Value before passing it to compute_val. This hides the details of CellVar such as index, and cache could use the index as the key.

katat: [a6ed103](https://github.com/zksecurity/noname/pull/69/commits/a6ed10355026028da43d73551935ccf08368c267)

src/constraints/field.rs
katat: ah, it shouldn't.
fix: a9e936e8546fcc2c445947b52ddca4af26d6ddd2

src/backends/mod.rs
katat: reverted to using usize: [2721919](https://github.com/zksecurity/noname/pull/69/commits/2721919849348ace25988f6346b470b69ce7d86d)

src/backends/kimchi/mod.rs
```diff
@@ -237,21 +238,36 @@ impl KimchiVesta {
...
     type Field = VestaField;
+    type CellVar = KimchiCellVar;
     type GeneratedWitness = GeneratedWitness;
 
     fn poseidon() -> crate::imports::FnHandle<Self> {
         builtin::poseidon
     }
 
-    fn witness_vars(&self, var: CellVar) -> &Value<Self> {
+    fn witness_vars(&self, var: KimchiCellVar) -> &Value<Self> {
```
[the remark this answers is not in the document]
katat: removed d9ed955c147e1500e2aad9f2ebb7c58d9f865947

src/constraints/field.rs
katat: That will be nicer.
I will implement https://github.com/zksecurity/noname/issues/56 in the [optimization PR](https://github.com/zksecurity/noname/pull/64) which can also wrap the constant as CellVar so that this new `sub` function can work with constant as well.

src/constraints/field.rs
katat: the comment 2 and 3 are connected

src/constraints/field.rs
katat: 674805e5

src/constraints/field.rs
```diff
@@ -256,15 +256,16 @@ pub fn if_else_inner<B: Backend>(
...
+    match cond {
         ConstOrCell::Const(cond) => {
             if cond.is_one() {
                 return Var::new_cvar(*then_, span);
             } else {
                 return Var::new_cvar(*else_, span);
             }
         }
-        ConstOrCell::Cell(_) => cond,
+        // skip
```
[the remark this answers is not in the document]
mimoo: another way to write this with less indentations:

```rust
if let ConstOrCell:Const(cond) = cond {
  ...
}
```

src/backends/kimchi/mod.rs
```diff
@@ -266,10 +266,6 @@ impl Backend for KimchiVesta {
         builtin::poseidon
     }
 
-    fn witness_vars(&self, var: KimchiCellVar) -> &Value<Self> {
-        self.witness_vars.get(&var.index).unwrap()
-    }
```
mimoo, line 250: nice :)

mimoo (approved):
nice!

src/var.rs
```diff
@@ -13,7 +13,7 @@ use crate::{
 };
 
 /// The signature of a hint function
-pub type HintFn<B: Backend> = dyn Fn(&B, &mut WitnessEnv<B::Field, B::CellVar>) -> Result<B::Field>;
+pub type HintFn<B: Backend> = dyn Fn(&B, &mut WitnessEnv<B::Field>) -> Result<B::Field>;
```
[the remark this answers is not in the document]
katat: yeah. I am wondering if it is possible to do poseidon for the kimchi part without using closure? (all via arithmetic)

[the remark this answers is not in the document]
mimoo: the answer is no because kimchi is a special gate :D so we need hints there, but I think it's enough to have a Value::Poseidon that knows exactly what the hint is (as opposed to pass it as a closure). Can you create an issue for that? I think it'd be nice for someone else to tackle :)

Issue #70, opened 2024-05-23
reorder processing of public inputs/outputs so that they are always before private inputs (important for r1cs)


mimoo:
The circuit writer will go through public inputs and private inputs in the order they're defined in the noname code, then public output.

Usually the public output needs to be placed first in proof system. The kimchi backend handles these things fine, because only the `add_public_input` creates a gate (and only the creation of gates create a new row in the symbolic witness table). Calling `new_internal_var` (which add_public_input and add_private input call) does not create a new row in the (symbolic) execution trace / witness table, because we don't know at this point if we are going to using them. 

In R1CS it's different, because the constraints are detached from the witness vector. So we do add the cells in the (symbolic) witness vector when we call `new_internal_var`.  We could technically wait to see if they are used in a constraint, but the compiler is supposed to warn the user if it doesn't use a variable, and even if they did that then it should be on them to pay the cost of adding useless variables in their code (but we should still warn them!). So I propose to keep the new_internal_var behavior and to add newly created variable to the witness vector right away.

So the important thing is to make sure we process the public input/output before the private inputs.

BTW, all these names are confusing as hell! The `witness_vars` map in the kimchi backend is actually just a map of variable to values (i.e. how to compute each variable, even if we never use them). (It could be a vector I think, although if we use deferred operations it might be good to keep it as a map.) But in r1cs we use `witness_vars` to mean the symbolic witness vector! In kimchi, the symbolic witness _table_ (coz we don't use a vector) is called `rows_of_vars`! We should probably change all these names...

Note: what is a symbolic witness table or vector? At proving time, we know that the witness vector will look like that: `[v0, v6, v7, v10, etc.]`. They're just variables at this point, we don't actually know the values because for example v0 will be a public input, v6 will be a private input, v7 will be v6 + 3 * v0, etc.

Basically they are inputs, outputs, or intermediary values of our circuits (so we don't know them yet). That's why we call them symbolic, coz they're not values, they're just symbols at this point. At runtime, then you replace the symbols with the actual values. [v0, v6, v7, v10, etc.] becomes [3, 5, 18, 3, etc.], if v0=3, and v6=5, etc.

So to summarize:

- [x] in generate_circuit we'll add `backend.init_circuit(&function.sig)?;` right before going through the arguments
- [x] in kimchi it won't do anything, but in r1cs it should add the constant `1` in witness_vars.
- [x] arguments should be sorted so that `add_public_input` and `add_public_output` are called BEFORE any `add_private_input`
[...]

PR #78, opened 2024-05-30
Reordering public input and outputs before private inputs

katat:
https://github.com/zksecurity/noname/issues/70


src/backends/r1cs/mod.rs
```diff
@@ -310,6 +311,21 @@ where
...
+    fn init_circuit(&mut self, func: &mut FunctionDef) {
+        // reorder the arguments by public and private
+        // the public inputs should be at the beginning
+        func.sig.arguments.sort_by_key(|arg| match &arg.attribute {
+            Some(attr) => match attr.kind {
+                crate::parser::types::AttributeKind::Pub => 0,
+                _ => 1,
+            },
+            None => 1,
+        });
```
mimoo, line 323: imo this should happen outside of this function, unless it becomes a backend specific thing for now this code could leave outside

src/backends/r1cs/mod.rs
```diff
@@ -374,6 +390,11 @@ where
 
         // check if every cell vars end up being a cell var in the circuit or public output
         for (index, _) in self.witness_vars.iter().enumerate() {
+            // skip the first var that is assumed to be the factor of constants of all linear combinations
```
mimoo, line 393: wdym "the factor of constants of all linear combinations"?

src/circuit_writer/mod.rs
```diff
@@ -150,11 +150,13 @@ impl<B: Backend> CircuitWriter<B> {
         let qualified = FullyQualified::local("main".to_string());
         let main_fn_info = circuit_writer.main_info()?;
 
-        let function = match &main_fn_info.kind {
+        let mut function = match &main_fn_info.kind {
             crate::imports::FnKind::BuiltIn(_, _) => unreachable!(),
             crate::imports::FnKind::Native(fn_sig) => fn_sig.clone(),
         };
```
mimoo, line 156: maybe clearer to write something like this:

```rust
// handle public inputs in order
for arg in function.sig.arguments.iter().filter(|arg| arg.is_public()) {
  handle_arg(arg, circuit_writer.add_public_inputs)?;
}

// handle private inputs in order
for arg in function.sig.arguments.iter().filter(|arg| arg.is_private())
  handle_arg(arg, circuit_writer.add_private_inputs)?;
}
```

which makes the handling of public inputs before private inputs much more explicit

examples/fixture/asm/r1cs/public_output.asm
```diff
@@ -1,4 +1,4 @@
 @ noname.0.7.0
 
-2 == (v_1 + v_2) * (1)
-v_1 + v_2 + 6 == (v_3) * (1)
+2 == (v_2 + v_3) * (1)
+v_2 + v_3 + 6 == (v_1) * (1)
```
mimoo, line 4: this is really nice to read to review these kind of PRs :) looks correct to me

book/src/cellvar.md
```diff
@@ -22,7 +22,7 @@ pub fn new_internal_var(&mut self, val: Value, span: Span) -> CellVar {
     self.next_variable += 1;
 
     // store it in the compiler
-    self.witness_vars.insert(var, val);
+    self.vars_to_value.insert(var, val);
```
mimoo, line 25: :') ❤️

src/backends/mod.rs
```diff
@@ -129,7 +129,7 @@ pub trait Backend: Clone {
 
     /// Compute the value of the symbolic cell variables.
     /// It recursively does the computation down the stream until it is not a symbolic variable.
-    /// - The symbolic variables are stored in the witness_vars.
+    /// - The cell vars are stored in the vars_to_value.
```
mimoo, line 132: maybe remove that comment since this is a backend thing now

book/src/witness-generation.md
```diff
@@ -16,17 +16,17 @@ pub struct Compiler {
     // ...
 
     /// This is used to compute the witness row by row.
-    pub rows_of_vars: Vec<Vec<Option<CellVar>>>,
+    pub witness_table: Vec<Vec<Option<CellVar>>>,
```
mimoo, line 19: ❤️

src/backends/r1cs/mod.rs
```diff
@@ -215,7 +215,7 @@ where
 {
     /// Constraints in the r1cs.
     constraints: Vec<Constraint<F>>,
-    witness_vars: Vec<Value<Self>>,
+    witness_vector: Vec<Value<Self>>,
```
mimoo, line 218: this is awesome \o/

src/backends/r1cs/snarkjs.rs
```diff
@@ -357,93 +294,3 @@ impl WitnessWriter {
...
-                    let new_index = snarkjs_exporter
-                        .witness_map
-                        .get(&original_var.index)
-                        .unwrap();
-                    assert!(rterms.contains_key(new_index));
-                }
-            }
-        }
-    }
-}
```
mimoo, line 449: nice to see how much code we're removing through this PR :D

mimoo (approved):
woop woop!

src/backends/r1cs/mod.rs
katat: The first var is assumed to be the coefficient of the constant of a linear combination.

In a linear combination, each of the vars can be paired with a coefficient.
The constant acts as the coefficient while var part is the always the first var (1). Look at it another way, the first var can always be the factor of the constants.

src/backends/r1cs/snarkjs.rs
katat: yeah, it is indeed simplify a lot things. Thanks for the suggestions!

src/circuit_writer/mod.rs
katat: Refactored: [99f328f](https://github.com/zksecurity/noname/pull/78/commits/99f328fa9d6799148eff5dd65a699758aeeceebe)

Thanks for the advice.

src/backends/r1cs/mod.rs
katat: [99f328f](https://github.com/zksecurity/noname/pull/78/commits/99f328fa9d6799148eff5dd65a699758aeeceebe)

src/circuit_writer/mod.rs
```diff
@@ -57,6 +57,9 @@ pub struct DebugInfo {
     pub note: String,
 }
 
+type HandleInput<B: Backend> =
```
mimoo, line 60: (imo not worth creating a type that's only used in one place)

src/backends/r1cs/mod.rs
mimoo: don't tell me, tell it in the comment :) I was just pointing out that the comment was not clear

Issue #77, opened 2024-05-30
create test for public output that could be linear combinations (in R1CS)


mimoo:
<img width="721" alt="Screenshot 2024-05-29 at 8 31 46 PM" src="https://github.com/zksecurity/noname/assets/1316043/e1f404b0-3e22-4393-9216-0a75deff15ba">

PR #79, opened 2024-05-30
add test for checking direct return from linear combination of inputs

katat:
https://github.com/zksecurity/noname/issues/77

It passes the test. Here is the reasoning:
- the public var is always added as a single value var at the beginning.
- the circuit writer keeps these public vars as they are.
- so the public vars can be recovered to a single cell var from linear combination.
- the public output var and return var are two different vars (although there is an `assert_eq` constraint among them at the end).
- the return var can be a linear combination of multiple vars, which is not an issue as they don't need to be converted to a single cell var.


examples/fixture/asm/r1cs/lc_return.asm
```diff
@@ -0,0 +1,3 @@
+@ noname.0.7.0
+
+v_1 + v_2 == (v_3) * (1)
```
mimoo, line 3: hopefully we'll see that with the reordering of public inputs it should be

```
v_1 + v_3 == v2
```

also what happened to write it like (A) * (B) = C ? it confuses me every time :D

mimoo (approved):
ohh, I guess the code I was looking at was about the public var, not the returned var!

examples/fixture/asm/r1cs/lc_return.asm
katat: currently, it is C = (A) * (B)

it is intended to align with the order of assignment in the noname code :D
so when it is in debug mode, this asm can align with the statement like
c = a + b

but this alignment might not be that useful though.

mimoo: I guess it's a matter of taste :o doesn't really matter too much

Issue #71, opened 2024-05-23
fix error when private input is unused


mimoo:
the following noname code:

```rust
fn main(private_input: Field, x2: Field, pub public_input: Field) {
    let xx = private_input + public_input;
    assert_eq(xx, 2);
}
```

turns into

```
   ╭─[examples/arithmetic.no:1:1]
 1 │ ╭─▶ fn main(private_input: Field, x2: Field, pub public_input: Field) {
 2 │ │       let xx = private_input + public_input;
 3 │ ├─▶     assert_eq(xx, 2);
   · ╰──── here
 4 │     }
   ╰────
  help: private input not used in the circuit
```

which is a weird error, it should highlight the `x2: Field` part and that's it

to reproduce, modify `arithmetic.no` and then:

```
cargo run test --backend kimchi-vesta --path examples/arithmetic.no --private-inputs '{"private_input": "1", "x2": "3"}' --public-inputs '{"public_input": "1"}' --debug
```

PR #87, opened 2024-06-03
fix: error when private input is unused

vuvoth:
#  Overview
Solve #71.  
I added span to private_input_indices and when if private input used by circuit, this will return error with private input span. 

# Result 
image

### DIFF
--- a/src/backends/r1cs/mod.rs
+++ b/src/backends/r1cs/mod.rs
@@ -220,7 +220,7 @@ where
     /// Record the public inputs for reordering the witness vector
     public_inputs: Vec<CellVar>,
     /// Record the private inputs for checking
-    private_input_indices: Vec<usize>,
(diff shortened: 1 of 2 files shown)


src/backends/r1cs/mod.rs
```diff
@@ -386,11 +386,15 @@ where
...
-                if self.private_input_indices.contains(&index) {
+                if let Some((_, private_input_span)) = self
+                    .private_input_indices
+                    .iter()
+                    .find(|(private_input_id, _)| private_input_id == &index)
+                {
                     let err = Error::new(
                         "constraint-finalization",
                         ErrorKind::PrivateInputNotUsed,
-                        main_span,
```
vuvoth, line 393: main_span not use anymore

katat (approved):
Nice, thanks a lot!

PR #91, opened 2024-06-03
chore(book): update assets for mdbook-admonish

eightfilms:
Fixes CI failure https://github.com/zksecurity/noname/actions/runs/9351865454/job/25738649882

Manually ran the command `mdbook-admonish install` found in the log and committed the changes.

### DIFF
--- a/book/book.toml
+++ b/book/book.toml
@@ -10,10 +10,10 @@ site-url = "/noname/"
 use-site-url-as-root = true
 curly-quotes = true
 git-repository-url = "https://www.github.com/mimoo/noname"
-additional-css = ["././mdbook-admonish.css"]
+additional-css = ["./mdbook-admonish.css"]
 
 [preprocessor]
 
 [preprocessor.admonish]
 command = "mdbook-admonish"
-assets_version = "2.0.0" # do not edit: managed by `mdbook-admonish install`
+assets_version = "3.0.2" # do not edit: managed by `mdbook-admonish install`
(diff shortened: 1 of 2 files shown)


book/book.toml
```diff
@@ -10,10 +10,10 @@ site-url = "/noname/"
 use-site-url-as-root = true
 curly-quotes = true
 git-repository-url = "https://www.github.com/mimoo/noname"
-additional-css = ["././mdbook-admonish.css"]
+additional-css = ["./mdbook-admonish.css"]
```
eightfilms, line 13: this was a change I did manually aside from running the command found in the failed job log - we previously seemed to be using a wrong path?

katat: Yeah, it looks like a wrong path.

katat:
Thanks!
I am wondering if we have to update these html files? They seem to be generated by the mdbook. @mimoo

eightfilms:
@katat Yeah it was auto-generated. I did a rough check by running `mdbook serve` locally and browsed a few pages to see if anything messed up and it doesn't appear to be, you might want to check too perhaps?

katat:
Ok, then we can exclude these html changes from this PR. I think they will be overwritten by the action builds.

eightfilms:
sure, let me fix this up.

eightfilms:
fixed!

katat:
Thanks so much!

PR #94, opened 2024-06-03
Fix example in docs/basics AND update examples/arithmetic.no

shankars99:
/basics had: `cargo run -- --path path/to/file.no --private-inputs '{"private_input": ["1"]}' --public-inputs '{"public_input": ["1"]}'` which was unrunnable as there was no Cmd provided. 

I updated it to use `test`. I also changed the example to take in `target: Field` as a parameter that checks if the sum of `public_input` and `private_input` add to the `target` value. 

I also changed `examples/arithmetic.no` to work with `target`.


katat:
Thanks for the PR!

It is true we'd better to use the `test` sub command. But I think we don't have to change the `arithmetic.no` code in order to update the readme, right?

shankars99:
I probably should have created a different PR for the arithmetic.no tbh. 

The default example of summing to equal to constant didn't seem too fun, and I figured that since it's arithmetic we could also have operations like mul and sub. 

Let me drop the arithmetic.no update from this PR and open a new one. Does that sound good?

PR #99, opened 2024-06-05
chore: improves tests negative

vuvoth:
# Changes 

- Make test more simple
- Use should panic macro

### DIFF
--- a/src/negative_tests.rs
+++ b/src/negative_tests.rs
@@ -14,7 +14,6 @@ macro_rules! create_test_with_error {
             let actual_error = res.unwrap_err().kind;
 
             assert!(matches!(actual_error, $expected_error));
-
         }
     };
 }
--- a/src/negative_tests.rs
+++ b/src/negative_tests.rs
@@ -1,23 +1,3 @@
-macro_rules! create_test_with_error {
-    ($test_name: ident, $expected_error: pat, $code: expr) => {
-        #[test]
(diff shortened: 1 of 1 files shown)


katat:
This simplifies the code a lot. But can we just use function instead of macro to achieve the same?

vuvoth:
> This simplifies the code a lot, thanks! But can we just use function instead of macro to achieve the same?

We can. However, we need create the function warps this function with #[test] macros for every test.

katat:
Hmm, yeah. But I think it is fine to use #[test] as it is a common practice for testing, and it is more explicit for code reading.

We are trying to avoid creating custom macros as much as possible, since they are generally harder to comprehend.

vuvoth:
Got it! So I will update PR. 
Btw I'm planing to add more tests for packages.

vuvoth:
Hi @katat! 
We can't pass something like ErrorKind::ReturnTypeMismatch(..) to function params. I also think the macro is not so hard to read in this case.

mimoo:
I agree with @katat here that macros should be sort of used as last case resort as they're easy to maintain/update, I don't think the changes are impactful enough here to warrant the introduction of a macro

vuvoth:
Hey @mimoo, @katat! 
Please review my PR again!

mimoo:
closing for now, feel free to reopen

PR #100, opened 2024-06-05
chore: fix all lints

eightfilms:
There's a lot of changes - I'll briefly summarize what I did:

I ran, as much as possible, the following:

```rust
cargo fix --all --allow-dirty
cargo fmt
cargo clippy --fix --all --allow--dirty
```

Some parts, I had to manually fix, or was out of scope of the current PR, so I added lint allowances + TODOs instead. An example of such 'out of scope' fixes include the `unused_must_use`s in `r1cs/snarkjs.rs` - these would require additional work on error handling which should probably belong in its own PR. For these cases, I prefer lint allowances since they allow us to easily find the spots where we want to fix later on.

Note that some of the lint allowances seem sane, so I didn't add TODOs for those, but only for those I felt would improve code quality if removed.

PR #105, opened 2024-06-06
chore: refactor `SnarkjsExporter` and `WitnessWriter` to handle errors

### LINKED ISSUE #102: properly handle errors in r1cs backend
#100 revealed that in [src/backends/r1cs/snarkjs.rs](https://github.com/zksecurity/noname/pull/100/files/a733b2f397ea7ee0dcc8f26638768abdf260fc40#diff-a62d19cfed9bfe61b9e9be5839be40f551f6fce4dea25c362dda1d2018e41446) that there are a few places where errors are left unhandled - these should not be silenced like in that PR, but should be properly handled.

Related comments:
https://github.com/zksecurity/noname/pull/100#discussion_r1628383632
https://github.com/zksecurity/noname/pull/100#discussion_r1628384126

eightfilms:
Closes #102 
Related: #100, addresses comments on unused results

Originally, all `Result`s were ignored. This PR updates the two structs/impls in question to: use either the appropriate `io::Result` or the standard `std::result::Result` (which the circom dep returns)

### DIFF
--- a/src/backends/r1cs/snarkjs.rs
+++ b/src/backends/r1cs/snarkjs.rs
@@ -316,6 +316,6 @@ impl WitnessWriter {
 
         let mut buffer = vec![0u8; size];
         buffer[..bytes.len()].copy_from_slice(&bytes);
-        Ok(self.inner.write_all(&buffer)?)
+        self.inner.write_all(&buffer)
     }
 }
--- a/src/backends/r1cs/snarkjs.rs
+++ b/src/backends/r1cs/snarkjs.rs
@@ -1,5 +1,7 @@
 use crate::backends::BackendField;
 use constraint_writers::r1cs_writer::{ConstraintSection, HeaderData, R1CSWriter};
+use miette::{miette, Diagnostic};
+use thiserror::Error;
 
 use std::collections::HashMap;
 use std::fs::{File, OpenOptions};
@@ -8,6 +10,20 @@ use std::vec;
 
 use super::{GeneratedWitness, LinearCombination, R1CS};
 use num_bigint_dig::BigInt;
+
+#[derive(Diagnostic, Debug, Error)]
+pub enum Error {
+    /// An error type associated with [`R1CSWriter`].
+    ///
+    /// It turns out that even [`R1CSWriter`] fails to write to file,
+    /// the error is ignored and a unit type is returned,
+    /// so we come up with a custom type to represent that.
+    #[error("Something went wrong writing the R1CS file")]
+    R1CSWriterIo,
+    /// An error type associated with [`WitnessWriter`].
+    #[error(transparent)]
+    WitnessWriterIo(#[from] std::io::Error),
+}
 
 #[derive(Debug)]
 struct SnarkjsConstraint {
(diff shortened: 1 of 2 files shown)


src/backends/r1cs/snarkjs.rs
```diff
@@ -1,10 +1,9 @@
 use crate::backends::BackendField;
-use crate::error::Result;
```
eightfilms, line 2: This is a bit awkward because it seems like the errors in this crate is intended to handle the DSL's errors, but there's no concept of a span in an IO error. Maybe a different error needs to be defined?

katat: hmm, interesting.

Do you think this can be a workaround if it is not related to DSL source code error? --> https://github.com/zkat/miette/blob/main/README.md#-in-application-code

katat:
Nice!

I think this is a good chance for us to figure out how to get these IO errors bubble up and be compatible with the miette::result at the top.

src/backends/r1cs/snarkjs.rs
eightfilms: I'd imagine we might find it useful to define dedicated `Error`s for each backend eventually, since IMO it might be weird to mixup DSL errors with backend errors.

In https://github.com/zksecurity/noname/pull/105/commits/28f1fb3906725005fe7dbf0842bd460f91c1cd73 I defined a new `Error` dedicated to snarkjs, perhaps we could use this as a start - What do you think?

src/cli/cmd_build_and_check.rs
```diff
@@ -428,9 +428,9 @@ where
     let r1cs_output_path = curr_dir.join("output.r1cs");
     let wtns_output_path = curr_dir.join("output.wtns");
 
-    snarkjs_exporter.gen_r1cs_file(&r1cs_output_path.clone().into_string());
+    snarkjs_exporter.gen_r1cs_file(&r1cs_output_path.clone().into_string())?;
```
eightfilms, line 431: we're now using the errors here.

src/backends/r1cs/snarkjs.rs
```diff
@@ -229,68 +247,75 @@ impl WitnessWriter {
...
     }
 
     /// Write a BigInt to the file
-    fn write_big_int(&mut self, value: BigInt, size: usize) {
+    fn write_big_int(&mut self, value: BigInt, size: usize) -> Result<(), io::Error> {
         let bytes = value.to_bytes_le().1;
 
         let mut buffer = vec![0u8; size];
         buffer[..bytes.len()].copy_from_slice(&bytes);
-        self.inner.write_all(&buffer);
```
katat, line 294: This can be just:

```
self.inner.write_all(&buffer)
```

src/backends/r1cs/snarkjs.rs
```diff
@@ -1,15 +1,30 @@
...
+    /// An error type associated with [`R1CSWriter`].
+    ///
+    /// It turns out that even [`R1CSWriter`] fails to write to file,
+    /// the error is ignored and a unit type is returned,
+    /// so we come up with a custom type to represent that.
+    #[error("Something went wrong writing the R1CS file")]
+    R1CSWriterIo,
+    /// An error type associated with [`WitnessWriter`].
+    #[error(transparent)]
+    WitnessWriterIo(#[from] std::io::Error),
```
katat, line 25: How does this error type work?

katat (approved):
Looks good to me. Just a minor issue mentioned above ^

src/backends/r1cs/snarkjs.rs
katat: That makes sense.

src/backends/r1cs/snarkjs.rs
katat: nvm

src/backends/r1cs/snarkjs.rs
eightfilms: fixed in [ea2564a](https://github.com/zksecurity/noname/pull/105/commits/ea2564a9a7f3446f2db56f2fbb94733896b32c98)

PR #106, opened 2024-06-07
chore: make CellVar index private

### LINKED ISSUE #39: make `private_input_indices` track `CellVar`s
I think `private_input_indices` should keep track of `CellVar`s instead of indices, as indices is a backend implementation detail.

Maybe we should remove the `pub` in `CellVar { pub index: usize` and see what breaks?

lognorman20:
Closes #39 

- Makes `CellVar` index private
- Refactor `private_input_indices` to `private_input_cell_vars` as a vector of `CellVar`s instead of indices

### DIFF
--- a/src/backends/kimchi/mod.rs
+++ b/src/backends/kimchi/mod.rs
@@ -98,7 +98,7 @@ pub struct KimchiVesta {
 
     /// Indexes used by the private inputs
     /// (this is useful to check that they appear in the circuit)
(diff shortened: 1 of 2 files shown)


katat:
Thanks a lot!
Maybe we can also apply the similar changes to the kimchi backend in this PR? It is the same issue over there.

lognorman20:
@katat thanks for the response, updated!

katat (approved):
Nice, thanks so much!

PR #107, opened 2024-06-08
update: docs addition example in /basics

shankars99:
1. made the adder example more interactive using a `target`
2. updated script to use `--test`

### DIFF
--- a/book/src/basics.md
+++ b/book/src/basics.md
@@ -5,9 +5,9 @@ Noname is a language that closely resembles Rust.
 For example, in the following program you can see a `main` function:
 
 ```rust
-fn main(pub public_input: Field, private_input: Field) {
-    let x = private_input + public_input;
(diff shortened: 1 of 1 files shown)


book/src/basics.md
```diff
@@ -5,9 +5,9 @@ Noname is a language that closely resembles Rust.
 For example, in the following program you can see a `main` function:
 
 ```rust
-fn main(pub public_input: Field, private_input: Field) {
-    let x = private_input + public_input;
-    assert_eq(x, 2);
+fn main(pub public_input: Field, private_input: Field, pub target: Field) {
+    let xx = private_input + public_input;
+    assert_eq(xx, target);
```
mimoo, line 10: not sure I understand why this is a good change :o

mimoo:
closing this one for now! feel free to reopen if you think this is still a good idea

PR #108, opened 2024-06-08
feat(types): impl `PartialEq` and `Hash` for `Ident` instead of `ModulePath`

eightfilms:
Extracted from #100 

This follows the original implementation. 2 `Ident`s should be equal if their values are equal, even if their `Span`s are different. This also includes manually implementing `Hash` to allow for this.

### DIFF
--- a/src/parser/types.rs
+++ b/src/parser/types.rs
@@ -156,7 +156,6 @@ pub struct Ty {
 }
 
 /// The module preceding structs, functions, or variables.
-// TODO: Hash should probably be implemented manually, right now two alias might have different span and so this will give different hashes
 #[derive(Default, Debug, Clone, Serialize, Deserialize, Hash, PartialEq, Eq)]
 pub enum ModulePath {
     #[default]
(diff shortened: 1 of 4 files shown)


src/parser/types.rs
```diff
@@ -401,12 +392,25 @@ impl FnSig {
...
+impl PartialEq for Ident {
+    fn eq(&self, other: &Self) -> bool {
+        self.value == other.value
+    }
+}
+
+impl Hash for Ident {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.value.hash(state);
+        self.span.hash(state);
```
mimoo, line 410: I think the TODO was explicitly about not having the span in the hash :o

mimoo: btw I think this solution would be cleaner: https://users.rust-lang.org/t/rust-crate-for-ignore-this-field-for-eq-partialeq-hash/33651/6

eightfilms: Hmm I see, I misunderstood the intention. I can fix it in this PR (or restore the TODO if that's preferable).

I think the `derivative` crate is [unmaintained](https://github.com/mcarton/rust-derivative/issues/117) and still uses syn v1, so that may not be a great idea. However an alternative ([educe](https://github.com/magiclen/educe)) was also presented in that issue, and that uses syn v2. We can bring that in, if this is preferable over manual implementation(s).

eightfilms: Using `educe` in [dfc03ad](https://github.com/zksecurity/noname/pull/108/commits/dfc03ad2a8e00558122e236de4f1ddb44cb652c5)

mimoo (approved):
awesome thanks a lot!

PR #109, opened 2024-06-08
chore: fix all unused_variables lint

### LINKED ISSUE #101: sort out unused variables
On fixing the codebase for clippy, #100 surfaced a few places where the compiler complains about unused variables:

```console
$ rg unused_var           
src/circuit_writer/mod.rs
3:#![allow(unused_variables)]

src/backends/kimchi/mod.rs
314:    #[allow(unused_variables)]

src/backends/r1cs/mod.rs
3:#![allow(unused_variables)]
```

We should sort out these unused variables, since they might be bugs. See [this comment](https://github.com/zksecurity/noname/pull/100#discussion_r1628382522)

eightfilms:
Closes #101 
Extracted from #100 

As the title says - this ended up being a mostly harmless change since most unused variables here seem to be either unused because of unimplemented features or unused due to trait requirements

### DIFF
--- a/src/backends/kimchi/mod.rs
+++ b/src/backends/kimchi/mod.rs
@@ -306,7 +306,7 @@ impl Backend for KimchiVesta {
         &mut self,
         public_output: Option<Var<Self::Field, Self::Var>>,
         returned_cells: Option<Vec<KimchiCellVar>>,
-        main_span: Span,
+        _main_span: Span,
     ) -> Result<()> {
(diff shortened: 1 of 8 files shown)


mimoo:
btw I'm not convinced that these are harmless, we should have a serious pass to ensure that we don't need to perform any checks on these variables in the typechecker (for example)

mimoo:
hey sorry but let's pun on this for now, as I think there were too many potentially dangerous changes :o

PR #110, opened 2024-06-09
Feat: noname cli: read input/output JSON from file

ibkisthere:
Closes #92 
There's a lot of changes - i'll briefly summarize what i did  

- I added a way to parse JSON files to the noname-cli , following the second option on #92 , it first tries to parse json, if it can not we will try parse as file path. 
- if the path/ file is invalid (it does not exists, or it is a directory) it panics immediately and shows an error message on the console 
- added 2 more Error types to the ParsingError Enum for the invalid path/ file does not exist error and path is directory
-  added an inputs folder and two folders under it for testing valid and invalid json public_inputs and private_inputs 

Here are some example commands to test the new changes to the cli

```cargo run test --backend r1cs-bn254 --path examples/arithmetic.no --private-inputs '{"private_input": "2"}' --public-inputs '{"public_input": "2"}' --debug```

``` cargo run test --backend r1cs-bn254 --path examples/arithmetic.no --private-inputs examples/json/inputs1/private_input.json --public-inputs examples/json/inputs1/public_input.json --debug```

PR #111, opened 2024-06-10
chore(cli): read input json files and strings

### LINKED ISSUE #92: noname cli: read input/output JSON from file
# Motivation 
We only read public input/output by passed in the cli as a string. This is good if we can read input from file. 

```bash
 cargo run test --backend r1cs-bn254 --path examples/arithmetic.no --private-inputs '{"private_input": "1"}' --public-inputs '{"public_input": "1"}' --debug
```
We can make cli above more simple: 

```bash
 cargo run test --backend r1cs-bn254 --path examples/arithmetic.no --private-inputs private_input.json --public-inputs ./public_inputs.json--debug 
```

# How to solve it?  
- Add option to read json file from cli.
- Rename --private-inputs, --public-inputs to --private-inputs-str... 

Another solution is try parse json, if can not we will try parse as file path.

lognorman20:
Closes #92. Allows user to pass either a JSON path or a JSON literal without a change to the current user frontend by expanding capabilities of the `parse_inputs` function

### DIFF
--- a/src/inputs.rs
+++ b/src/inputs.rs
@@ -1,6 +1,6 @@
 //! Used to parse public and private inputs to a program.
 
-use std::{collections::HashMap, str::FromStr};
+use std::{collections::HashMap, fs::File, io::Read, str::FromStr};
 
 use ark_ff::{One, PrimeField, Zero};
 use miette::Diagnostic;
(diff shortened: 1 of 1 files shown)


katat:
Thanks. 

I think we can add a new function, such as `read_json_file`, which does 
1. read the file content
2. pass the content to the current function parse_inputs

Then update the command line logic to determine which function to call: either `read_json_file` or `parse_inputs`.

So instead of making a single function to accept two types of inputs, this may make the code clearer.

mimoo:
ah, same as this one right? https://github.com/zksecurity/noname/pull/110 I guess we have to decide which approach to take now :D

mimoo:
let's merge this one, it looks simple/clean enough and if it works then it's great to have that option!

PR #112, opened 2024-06-10
chore(cli): cleanup backend option

eightfilms:
derive `ValueEnum` for `BackendOpt` - we use `clap`'s builtin enum support to do validation and parsing.

Also implement `From<BackendOpt> for BackendKind` to simplify the flow.

Some screenshots:

Not specifying any backends: `cargo run -- run --backend`

Screenshot 2024-06-10 at 9 20 57 PM

Mistyping a backend: `cargo run -- run --backend kimchi`

Screenshot 2024-06-10 at 9 21 17 PM

PR #115, opened 2024-06-10
r1cs is default backend, default example works

mimoo:
the default example didn't work with R1CS. Also I made R1CS the default so that we don't have to pass `--backend` by default.


katat (approved):

eightfilms:
`run` still fails with a `unwrap` on a `None` if `--backend` isn't passed btw so that should probably be fixed. Encountered it while going through the book: https://zksecurity.github.io/noname/basics.html.

#112 semi-fixes it by making the argument not an `Option`

PR #118, opened 2024-06-11
Replace gate creation by wiring for copy constraints

levihackerman-102:
I've removed the gate creation in ```assert_eq_var()```. I will be testing it further, for now I'm opening a draft PR.


levihackerman-102:
I think it would be better to create a method for directly adding a copy constraint for permutation then I can use it in both ```assert_eq_var()``` and ```assert_eq_const()```.

levihackerman-102:
I'm getting some errors in the Kimchi backend tests when I run them:
```
failures:

---- backends::kimchi::prover::tests::test_public_output_constraint stdout ----
thread 'backends::kimchi::prover::tests::test_public_output_constraint' panicked at src/backends/kimchi/prover.rs:93:26:
index out of bounds: the len is 4 but the index is 4
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

---- tests::examples::test_bool::case_1_kimchi_vesta stdout ----
thread 'tests::examples::test_bool::case_1_kimchi_vesta' panicked at src/backends/kimchi/prover.rs:93:26:
index out of bounds: the len is 7 but the index is 7
```

I'll try debugging them.

levihackerman-102:
I'm getting some out of bound errors when running the cargo tests. It seems to be an off-by-one error but I'm not sure what caused it.

levihackerman-102:
@mimoo @katat do you have some idea of what might be going wrong here?

katat:
Hi @levihackerman-102 
I am not quite sure what is this PR for. Can you create an issue to describe what this PR is aiming to achieve?

levihackerman-102:
> Hi @levihackerman-102 I am not quite sure what is this PR for. Can you create an issue to describe what this PR is aiming to achieve?

@katat this is for issue #55.

src/backends/kimchi/mod.rs
```diff
@@ -704,7 +762,7 @@ impl Backend for KimchiVesta {
 
     fn assert_eq_var(&mut self, lhs: &KimchiCellVar, rhs: &KimchiCellVar, span: Span) {
         // TODO: use permutation to check that
-        self.add_generic_gate(
+        self.add_copy_constraint(
```
mimoo, line 765: you should be able to modify the wiring directly (in the same way `add_gate` adds information to the wiring, I'm not sure what is happening here :D

levihackerman-102: Oh, I thought of doing it in a function so that we could call it in both the ```assert()``` calls, I'll do it directly instead then.

levihackerman-102: I've added the wire modification directly in the functions. The tests are still failing though:
```
---- tests::examples::test_equals::case_1_kimchi_vesta stdout ----
thread 'tests::examples::test_equals::case_1_kimchi_vesta' panicked at src/backends/kimchi/prover.rs:99:26:
index out of bounds: the len is 15 but the index is 15

---- tests::examples::test_arithmetic::case_1_kimchi_vesta stdout ----
thread 'tests::examples::test_arithmetic::case_1_kimchi_vesta' panicked at src/backends/kimchi/prover.rs:99:26:
index out of bounds: the len is 3 but the index is 3

---- tests::examples::test_lc_return::case_1_kimchi_vesta stdout ----
thread 'tests::examples::test_lc_return::case_1_kimchi_vesta' panicked at src/backends/kimchi/prover.rs:99:26:
index out of bounds: the len is 3 but the index is 3

---- tests::examples::test_iterate::case_1_kimchi_vesta stdout ----
thread 'tests::examples::test_iterate::case_1_kimchi_vesta' panicked at src/backends/kimchi/prover.rs:99:26:
index out of bounds: the len is 4 but the index is 4

---- tests::examples::test_sudoku::case_1_kimchi_vesta stdout ----
thread 'tests::examples::test_sudoku::case_1_kimchi_vesta' panicked at src/backends/kimchi/prover.rs:99:26:
index out of bounds: the len is 22150 but the index is 22150
```

I'm getting out of bound errors. What could be the reason for this? I thought that maybe it's checking the number of copy constraints with gates and I'm not adding gates anymore so maybe that's why there's an error.

src/backends/kimchi/mod.rs
```diff
@@ -688,28 +688,105 @@ impl Backend for KimchiVesta {
...
+            cst.neg(),
+        ];
+        let note = "constrain var - cst = 0 to check equality";
+
+        // sanitize
+        assert!(coeffs.len() <= NUM_REGISTERS);
+        assert!(vars.len() <= NUM_REGISTERS);
+
+        // construct the execution trace with vars, for the witness generation
+        self.witness_table.push(vars.clone());
```
mimoo, line 707: you are using the witness table here, where we should not add anything to the witness table as we just want to wire things around. (Also, note to myself, variables are always constrained in circuit at this point? maybe we should differentiate between variable and allocated variable, cc @katat )

levihackerman-102: I've removed the witness table additon.

mimoo:
so I don't think the approach is right. Now that I think about it I think we'll need something like union finds to make it work which is a bit more involved. I'll update the issue with more information

levihackerman-102:
> so I don't think the approach is right. Now that I think about it I think we'll need something like union finds to make it work which is a bit more involved. I'll update the issue with more information

Alright, I'm willing to continue on this issue, are there any parts of the documentation that I could read to find out why this approach wouldn't work and what we can do to fix it? I'll also pickup another issue in the meantime.

mimoo:
I wrote more about it here, let me know if that helps! https://github.com/zksecurity/noname/issues/55#issuecomment-2171904782

mimoo:
I'll close this one, but feel free to reopen later!

PR #119, opened 2024-06-12
Add support for the not equal operator

### LINKED ISSUE #113: Support not equal operator in noname
# Motivation

We should support not equal operator in noname. 

```
assert(a != b)
``` 

# Solution

lognorman20:
Adds the `!=` operator for variables in Noname, closes #113. 

Summary of changes:
* Implemented `is_zero_cell` to check if a given cell is equal to zero. This function closely follows the [Gnark implementation](https://github.com/Consensys/gnark/blob/master/frontend/cs/r1cs/api.go#L535). I'm imagining this can be further simplified to reduce the total number of constraints necessary -- perhaps taking into account the [Circom implementation](https://github.com/iden3/circomlib/blob/cff5ab6288b55ef23602221694a6a38a0239dcc0/circuits/comparators.circom#L24).
* Implement `not_equal` to check if two given cells are equal. The check is done by calculating the difference between the two and checking if it's zero
* Update the lexer, parser, writer, and type checker to recognize `!=`
* Add example of using `!=` in `examples/not_equal.no`
* Add a test for the `!=` operator


src/constraints/field.rs
```diff
@@ -220,6 +220,81 @@ fn equal_cells<B: Backend>(
...
+
+            // a * m = 0 -- constrain m to be 0 if a != 0
+            let a_mul_m = compiler.backend.mul(&a, &m, span);
+
+            compiler.backend.assert_eq_const(&a_mul_m, zero, span);
+
+            Var::new_var(m, span)
+        }
+    }
+}
```
mimoo, line 296: I need to have a second pass here, but why not using `equal_cells`?

examples/fixture/asm/r1cs/not_equal.asm
```diff
@@ -0,0 +1,12 @@
...
+v_4 == (v_1 + -1 * v_2) * (v_3)
+v_5 + -1 == (-1 * v_4) * (1)
+v_6 == (v_1 + -1 * v_2) * (v_5)
+0 == (v_6) * (1)
+1 == (-1 * v_5 + 1) * (1)
+v_8 == (v_1 + -7) * (v_7)
+v_9 + -1 == (-1 * v_8) * (1)
+v_10 == (v_1 + -7) * (v_9)
+0 == (v_10) * (1)
+1 == (-1 * v_9 + 1) * (1)
```
mimoo, line 12: so many constraints :o I'm wondering if we're missing good optimizations here

lognorman20: i agree there definitely can be some optimization, i'm imagining in the `is_zero_circuit`?

src/constraints/field.rs
lognorman20: This is all based on [this comment](https://github.com/zksecurity/noname/issues/113#issuecomment-2159259007). Where do you imagine using `equal_cells`? My first thought was to do

```
let diff = sub(compiler, &lhs[0], &rhs[0], span);
let res = equal_cells(compiler, &diff, &zero, span);
```

But that doesn't use the `is_zero` operator mentioned

mimoo: do you wanna try another PR with this to see how the constraints differ :D?

src/constraints/field.rs
```diff
@@ -220,6 +220,81 @@ fn equal_cells<B: Backend>(
...
+                .backend
+                .new_internal_var(Value::Inverse(a.clone()), span);
+
+            // m = -a*x + 1 -- constrain m to be 1 if a == 0
+            let ax = compiler.backend.mul(&a, &x, span);
+            let neg_ax = compiler.backend.neg(&ax, span);
+            let m = compiler.backend.new_internal_var(
+                Value::LinearCombination(vec![(one, neg_ax.clone())], one),
+                span,
+            );
```
mimoo, line 283: you should be able to write `compiler.backend.sub(Var::new_constant(one, span) , &ax, span);` no? (cc @katat )

mimoo:
code looks good to me, but as pointed out do you want to try first in another PR if the solution using `equal_cell` leads to less constraints?

src/constraints/field.rs
lognorman20: sure!

lognorman20:
yes i can try to reduce constraints in another pr

lognorman20:
@mimoo addressed in #128!

mimoo:
maybe is_equal should use is_zero(a-b) now?

PR #120, opened 2024-06-12
Support default array declaration

katat:
This PR is to support const generic for default array declaration.

02bad12b62b0b325394f1877b812efca58ae0868 is the test cases to showcase this new feature.

Still figuring out how to compute the value behind a constant var that is out of context in a function: https://github.com/zksecurity/noname/blob/48ec1061b6a927c72e329d97342100a3050c5d75/src/circuit_writer/writer.rs#L668-L674


katat:
After supporting the const generic variable to declare a default array, we need a way for the [ArrayAccess](https://github.com/zksecurity/noname/blob/48ec1061b6a927c72e329d97342100a3050c5d75/src/circuit_writer/writer.rs#L638) to induce the value behind size variable to do sanity checks.

Take the following code for example, we need to know the sizes of the houses array and the rooms array in order to check if the array access is within their bounds:
```rust
const houses_num = 5;
const rooms_num = 3;
const room_size = 20;

struct Room {
    size: Field,
}

struct House {
    rooms: [Room; rooms_num],
}

fn build_house(const num: Field) -> [House; num] {
    let houses = [
        House {
            rooms: [Room {size: room_size}; rooms_num]
        }; 
        num
    ];
    return houses;
}

fn main(pub resize_to: Field, total_size: Field) {
    let mut sum_size = 0;
    let mut houses = build_house(houses_num);

    for ii in 0..houses_num {
        for jj in 0..rooms_num {
            houses[ii].rooms[jj].size = resize_to;
            sum_size = sum_size + houses[ii].rooms[jj].size;
        }
    }

    assert_eq(sum_size, total_size);
}
```

In the circuit generation process of circuit writer, these array sizes can be induced after the array declaration expressions are computed. Because these declaration expression node ids are unique, we could save the computed size value with the expression node id as the key in a map for later references.

For example, the following is the array declaration expression node for the field `rooms`:
```
[Room {size: room_size}; rooms_num]
```

Now the problem is how can we trace the variables' expression nodes, such as `houses[ii].rooms[jj]`, to their declaration expression nodes above ^. So we can look up the `rooms` array size by node id via the map stored earlier.

As we can [compute the type](https://github.com/zksecurity/noname/blob/e1ce80847ee381e4f4f2d364f409314c5004521c/src/type_checker/checker.rs#L99) for an expression node, I am wondering if it is currently possible (without major refactoring) to attach the declaration node when storing a [computed type](https://github.com/zksecurity/noname/blob/e1ce80847ee381e4f4f2d364f409314c5004521c/src/type_checker/checker.rs#L670).

katat:
I went with the `ComputedExpr` approach as described in the rfc draft [69033d3](https://github.com/zksecurity/noname/pull/120/commits/69033d35b97a012bfd304e07be40046ecd07dd8a)

mimoo (commented):
some quick comments, but can you cut the PR into commits so I can review it commit by commit? danke!

src/inputs.rs
```diff
@@ -76,8 +76,13 @@ impl<B: Backend> CompiledCircuit<B> {
...
             (TyKind::Array(el_typ, size), Value::Array(values)) => {
-                if values.len() != (*size as usize) {
-                    panic!("wrong size of array");
+                match size {
+                    ArraySize::Number(n) => {
+                        if values.len() != *n as usize {
+                            panic!("wrong size of array");
+                        }
+                    }
+                    ArraySize::ConstVar(_) => panic!("array argument with size in const var is not supported for main function"),
```
mimoo, line 85: in general we should return nicer error to the user if we can :D, here we could even return an error with a span

src/type_checker/mod.rs
```diff
@@ -126,7 +127,17 @@ impl<B: Backend> TypeChecker<B> {
                 sum
             }
             TyKind::BigInt => 1,
-            TyKind::Array(typ, len) => (*len as usize) * self.size_of(typ),
+            TyKind::Array(typ, size) => match size {
+                ArraySize::Number(n) => (*n as usize) * self.size_of(typ),
+                ArraySize::ConstVar(var) => {
```
mimoo, line 132: at this point I'm confused, I would have expected sizeof to either return a symbolic value, or an actual value. But here we manage to actually return an actual value. This should only work when we use something like:

```rust
const val = 4;
some_call([0; val]);
```

right? but not with const generics

src/type_checker/checker.rs
```diff
@@ -52,24 +53,44 @@ struct ExprTyInfo {
 
     /// The type of the expression node.
     typ: TyKind,
+
+    /// This is needed to pass around the const value (values not represented by variables)
+    /// that is not stored in the global constants hashmap.
+    /// For example:
+    /// let xx = [1; 10]; // where 10 is a const value that should be passed around in the type checker
+    value: Option<String>,
```
mimoo, line 61: interesting

mimoo: also isn't that information contained in `TyKind::Array(_, size)`?

katat: Oh right, this addition of `value` might be unnecessary. We just need `TyKind::Array(_, size)`

src/type_checker/mod.rs
katat: That is a nice catch. The `size_of` in type checker should only return symbolic value. Then in the circuit generation it should manage to return an actual value.

src/circuit_writer/mod.rs
```diff
@@ -234,18 +240,28 @@ impl<B: Backend> CircuitWriter<B> {
         // get length
         let len = match &typ.kind {
             TyKind::Field => 1,
-            TyKind::Array(typ, len) => {
+            // todo: the size will get from const var
```
mimoo, line 243: not sure I understand how actionable that todo is

src/circuit_writer/mod.rs
```diff
@@ -234,18 +240,28 @@ impl<B: Backend> CircuitWriter<B> {
...
                 if !matches!(**typ, TyKind::Field) {
                     unimplemented!();
                 }
-                *len as usize
+                // *size as usize
+                match size {
+                    ArraySize::Number(n) => *n as usize,
+                    ArraySize::ConstVar(_) => panic!(
+                        "array argument with size in const var is not supported for main function"
+                    ),
```
mimoo, line 253: needs a more friendly user error than a panic :o

book/src/rfc/const_generic.md
```diff
@@ -0,0 +1,146 @@
...
+
+### Introduce ComputedExpr
+
+With the const variable as array size, we need to propagate its value in circuit generation phase to check if the access is within the bound of the array. So we need a way to retain structure information of a variable that is currently represented by `VarOrRef`.
+
+We can add a layer of abstraction to the `VarOrRef` to keep track of the structure of the data. We can call it `ComputedExpr`. It is computed from `compute_expr` function, which recursively calculates the values (`cvars`) behind the `Expr`. Then the `ComputedExpr` keeps track of the structure of the data, and being passed around.
+
+With the `ComputedExpr` as the result of `compute_expr` calls, the field / array access can determine the structure of the value with actual size of the current variable. For example, when we are accessing `houses[1].rooms[2]`, it can check if the access is within the bound of `houses` and `rooms`, by checking the structure and size of a `ComputedExpr` representing the targeted access path.
+
+In simple view, `ComputedExpr` can be seen as an `Expr` with const variables folded into actual values.
```
mimoo, line 140: what is "simple view"?

mimoo: folded into -> replaced with?

book/src/rfc/const_generic.md
```diff
@@ -0,0 +1,146 @@
...
+
+## type checking const variable
+
+The current type checker is designed to do local type checks. For the cases of const variable as array size, the type checker has to trace the values globally. This is out of the scope for the design of the current type checker.
+
+Instead of checking the const variable in the type checker, we can check it in the circuit generation phase. 
+
+### changes from previous version
+
+In current circuit generation phase, it computes the low level variables `cvars` via the NAST. It starts with the main function and recursively allocates `cvars` to different variables `VarOrRef`. These `VarOrRef` are passing around in the circuit generation process for other different places to further calculate.
```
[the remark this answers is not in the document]
mimoo: cvars should be computed by circuit writer, not by NAST no?

src/circuit_writer/writer.rs
```diff
@@ -109,6 +113,308 @@ impl Ord for AnnotatedCell {
...
+
+impl<F: BackendField, V: BackendVar> Display for Access<F, V> {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        let mut s = self.var_name.clone();
+        for step in &self.steps {
+            match step {
+                AccessKind::Field(field) => s.push_str(&format!(".{}", field)),
+                AccessKind::Array(index) => s.push_str(&format!("[{}]", index)),
+            }
+        }
```
mimoo, line 169: can't you just write `self.steps.map(ToString::to_string).join()` if you have `Display` implemented on `AccessKind`?

src/circuit_writer/mod.rs
katat: If the ArraySize is a ConstVar, then it can load the const value. This is assume the const variable is in the local module. I think the `ArraySize::ConstVar` would need to be a something like `ExprKind::Variable` that can point to other module's constants if we can support that scenario.

book/src/rfc/const_generic.md
katat: I meant "in a sense" :D

src/type_checker/mod.rs
mimoo: only return symbolic value IF there's a const generic involved

src/circuit_writer/mod.rs
mimoo: (you should write that in the TODO ;))

src/type_checker/mod.rs
katat: That makes sense. The type checker shouldn't tracing the actual value for the const vars.

So we may need to change the function return
```rust
fn size_of(&self, typ: &TyKind) -> usize
```

to 

```rust
fn size_of(&self, typ: &TyKind) -> SymbolicValue
```

where `SymbolicValue` can represent something like:

```
cst * 10 + 5
```

mimoo: yeah that makes sense, something like:

```rust
enum Symbolic {
  Add(Symbolic, Symbolic),
  Mul(Symbolic, Symbolic),
  Concrete(BigInt)
}
```

katat:
implemented by another approach: https://github.com/zksecurity/noname/pull/147

PR #122, opened 2024-06-12
Fix hex literal

### LINKED ISSUE #121: Hex literals don't work
Parser creates the `Hex` literal token (instead of creating `BigInt` or `Field` token) and lexer doesn't handle it at all.

MartinOndejka:
closes #121 

Gets rid of `Hex` token and instead immediately returns the `BigInt`.


MartinOndejka:
Maybe the BigInt token should already parse the BigInt instead of passing the String? Not sure about that.

mimoo (commented):
thanks for the PR! Can you actually create a test where a number is passed as a hex string `0xsomething`?

src/lexer/mod.rs
```diff
@@ -230,7 +230,12 @@ impl Token {
                     let token_type = if is_numeric(&ident_or_number) {
                         TokenKind::BigInt(ident_or_number)
                     } else if is_hexadecimal(&ident_or_number) {
-                        TokenKind::Hex(ident_or_number)
+                        let hex = ident_or_number.trim_start_matches("0x");
+                        TokenKind::BigInt(
+                            BigInt::from_str_radix(hex, 16)
+                                .expect("invalid hex number")
```
mimoo, line 236: I think we should return a proper error here

src/lexer/mod.rs
```diff
@@ -230,7 +230,12 @@ impl Token {
...
+                                .to_str_radix(10),
```
mimoo, line 237: what's weird is that we're doing some parsing here, but not on line 231. WDYT? Should we instead have TokenKind::Bigint enclose a BigInt type?

mimoo (approved):
looks good to me, but I just made a suggestion which could be pushed in this PR (or in another PR if you don't want to :D)

src/lexer/mod.rs
MartinOndejka: It shouldn't be possible to get an error here since `is_hexadecimal` would return false. Maybe return error if the number starts with `0x` and rest is not a valid hexadecimal value?

src/lexer/mod.rs
MartinOndejka: Yeah I think so, I can implement that

MartinOndejka: I've renamed also BigInt token/expression to BigUInt, since it gets signed with `-`

mimoo:
beautiful! thanks so much!

Issue #93, opened 2024-06-03
Type check bugs: compiler accept when we use function name as variable. 


vuvoth:
# Context 

```rust
fn main(pub public_input: Field, private_input: Field) {
    let double = private_input + public_input;
    assert_eq(double, 2);
}

fn double(xx: Field) -> Field {
    return xx + xx;
} 
```

A problem above compile success in noname. 

# How to fix it? 
We should fix type checker.

MohamedElqdusy:
would @mimoo assign me this issue, please?

MohamedElqdusy:
@vuvoth your example should be a valid Rust syntax, I think it should not compile only after `double()` call as the following 
```
fn main(pub public_input: Field, private_input: Field) {
    let double = private_input + public_input;
    assert_eq(double, double(private_input , public_input));
}

fn double(pub public_input: Field, private_input: Field) -> Field {
    return private_input + public_input;
} 
```

PR #125, opened 2024-06-15
Function name in use check

MohamedElqdusy:
Fix https://github.com/zksecurity/noname/issues/93

for the following example 
```
use std::crypto;

fn double(xxx: Field) -> Field {
    return xxx + xxx;
} 

fn main(pub public_input: Field, private_input: Field) {
    let double = private_input;
    assert_eq(double, double(private_input));
}

``` 
the compiler should return
```
Error:   × Looks like something went wrong in type-checker
   ╭─[../../examples/example.no:2:1]
 2 │ 
 3 │ fn double(xxx: Field) -> Field {
   ·       ─┬─
   ·          ╰── here
 4 │     return xxx + xxx;
   ╰────
  help: function name `double` is already in use by a variable present in the scope
```

I updated  the test after reflecting the new checks work by making the pipeline fail at https://github.com/zksecurity/noname/actions/runs/9528372535/job/26265972691?pr=125#step:6:324

### DIFF
--- a/src/type_checker/checker.rs
+++ b/src/type_checker/checker.rs
@@ -596,7 +596,7 @@ impl<B: Backend> TypeChecker<B> {
             StmtKind::Expr(expr) => {
                 // make sure the expression does not return any type
                 // (it's a statement expression, it should only work via side effect)
-
+                
                 let typ = self.compute_type(expr, typed_fn_env)?;
                 if typ.is_some() {
                     return Err(self.error(ErrorKind::UnusedReturnValue, expr.span));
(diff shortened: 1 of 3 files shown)


src/type_checker/checker.rs
```diff
@@ -623,6 +623,17 @@ impl<B: Backend> TypeChecker<B> {
...
+        match typed_fn_env.get_type_info(&fn_sig.name.value) {
+            Some(_) => {
+                return Err(self.error(
+                    ErrorKind::FunctionNameInUsebyVariable(fn_sig.name.value),
+                    fn_sig.name.span,
+                ))
+            }
+            None => (),
+        };
+
```
mimoo, line 636: imo we should do it the other way around, when we create a new variable (var assigment statement) we should check if a function with the same name doesn't already exist.

MohamedElqdusy: I think functions are being declared statically less than variables, **a function might contain many variables** but not vice versa, checking names on functions' declaration reduces our performance penalty for such extra checks.

mimoo: I guess that makes sense, in this case we're not checking that you CAN'T name a variable name after a function, but rather that if you do you don't end up calling the function with the same name too. It's a bit restrictive but I guess it would be confusing to write such code so I'm in for the compiler deny

mimoo: that makes sense!

mimoo (approved):
thanks for the PR!

PR #128, opened 2024-06-17
Not equal using `equal_cells`

lognorman20:
This PR aims to optimize #119. Implements `equal_cells` in `is_zero` in an effort to reduce constraints. The result of doing this are presented in this table:

| **Backend** | **Constrains w/o not_equal** | **Constraints w/ not_equal** |
|:-----------:|:----------------------------:|:----------------------------:|
|     R1CS    |              10              |              12              |
|    Kimchi   |              19              |              24              |

Although the number of constraints went in the wrong direction, I have an idea for improvement. First is implementing more bitwise operations to avoid adding constraints. Functions such as `equal`, `is_zero_cell`, and even arithmetic operations like `sub` and `mul` can be implemented using bitwise operations that avoid creating unnecessary internal variables. As the language continues to expand, more operations will come such as bitshifts and xor which will enhance the possibilities to optimize. Moving the workload to bitwise operators to reduce constraints might just be moving the problem to another area though and I'm curious to hear what other ideas there are.


lognorman20:
Another aspect I noticed is that the # of constraints for equal and not equal are roughly the same. Definitely some improvement can be done for both

src/constraints/field.rs
```diff
@@ -220,6 +220,61 @@ fn equal_cells<B: Backend>(
...
+        span,
+    );
+    let mut acc = Var::new_var(acc, span);
+
+    for (l, r) in lhs.cvars.iter().zip(&rhs.cvars) {
+        let diff = sub(compiler, l, r, span);
+        let res = is_zero_cell(compiler, &diff[0], span);
+        let not_res = boolean::not(compiler, &res[0], span);
+        acc = boolean::and(compiler, &not_res[0], &acc[0], span);
+    }
```
mimoo, line 252: btw there's probably a faster way to implement that, just sum the results of `is_zero_cell`, and then do a `not(is_zero(acc))` at the end

mimoo:
thanks for the experiment here! I can close this one as we merged the other!

PR #131, opened 2024-06-19
feat: Allow array of fields on public output

### LINKED ISSUE #130: Array as public output
Hi! I would like to be able to output an array of public values when writing a `noname` circuit. An example would be:

```rust
fn main(pub public_input: [Field; 2], other_private_input: Field) -> [Field; 2] {
        let x = private_input * public_input[0];
        let y = private_input * public_input[1];
       return [x, y]
};
```

Is this something that could be done? I'm targeting the R1CS backend.

erhant:
Here is my humble attempt at #130. I have added a test & it seems to compile & pass.

- Resolves #130
- Changes type of public outputs within `generate_witness` of Kimchi backend to use `BTreeMap` instead of `HashMap` to guarantee ordering at the final output


src/circuit_writer/mod.rs
```diff
@@ -163,12 +163,15 @@ impl<B: Backend> CircuitWriter<B> {
 
         // create public output
         if let Some(typ) = &function.sig.return_type {
-            if typ.kind != TyKind::Field {
-                unimplemented!();
+            match typ.kind {
+                TyKind::Field => {
+                    circuit_writer.add_public_outputs(1, typ.span);
+                }
+                TyKind::Array(_, len) => {
```
mimoo, line 170: I don't think this is correct as this won't work if you have `[T; len]` for `T` a different type than `Field`.

I think a quick fix here could be to match for `TyKind::Array(TyKind::Field, len)` or to actually recurse and figure out the real size (akin to the size_of function), but this would be out of the scope of this issue and perhaps a tad more complicated :o

erhant:
I tried the following circuit:

```rs
fn main(pub public_input: Bool, private_input: Bool) -> [Bool; 3] {
    // constants
    let xx = false && false;
    assert(!xx);

    // private input must be true
    let yy = private_input && true;
    assert(!yy);

    // public input must be false
    let zz = public_input && true;
    assert(zz);

    return [!yy, yy, zz];
}
```

with commands:

```sh
cargo run test -p examples/bool_output_array.no --public-inputs '{"public_input": true}' --private-inputs '{"private_input": false}' --backend kimchi-vesta

cargo run test -p examples/bool_output_array.no --public-inputs '{"public_input": true}' --private-inputs '{"private_input": false}' --backend r1cs-bls12-381
```

and it seems to compile fine! The outputs are:

```sh
@ noname.0.7.0

DoubleGeneric<1>
DoubleGeneric<1>
DoubleGeneric<1>
DoubleGeneric<1>
DoubleGeneric<1,0,-1,0,-1>
DoubleGeneric<0,0,-1,1>
DoubleGeneric<1>
DoubleGeneric<1,0,-1,0,-1>
DoubleGeneric<0,0,-1,1>
DoubleGeneric<1>
DoubleGeneric<1,1>
DoubleGeneric<1,0,-1,0,1>
DoubleGeneric<1,0,0,0,-1>
DoubleGeneric<1,0,0,0,-1>
DoubleGeneric<1,1>
DoubleGeneric<1,0,-1,0,1>
DoubleGeneric<1,-1>
DoubleGeneric<1,-1>
DoubleGeneric<1,-1>
(0,0) -> (16,0)
(1,0) -> (17,0)
(2,0) -> (18,0)
(3,0) -> (4,0) -> (5,0) -> (13,0) -> (18,1)
(4,2) -> (5,1)
(5,2) -> (6,0)
(7,0) -> (8,0) -> (10,0) -> (14,0) -> (17,1)
(7,2) -> (8,1)
(8,2) -> (9,0)
(10,1) -> (11,0)
(11,2) -> (12,0)
(14,1) -> (15,0)
(15,2) -> (16,1)
```

and

```sh
@ noname.0.7.0

v_5 == (v_4) * (v_4 + -1)
0 == (v_5) * (1)
v_7 == (v_6) * (v_6 + -1)
0 == (v_7) * (1)
1 == (-1 * v_6 + 1) * (1)
1 == (v_4) * (1)
-1 * v_6 + 1 == (v_1) * (1)
v_6 == (v_2) * (1)
v_4 == (v_3) * (1)
```

Will try with a custom type in a moment!

erhant:
```rs
struct Thing {
    xx: Field,
    yy: Field,
}

fn main(pub xx: Field, pub yy: Field) -> [Thing; 2] {
    let thing1 = Thing {
        xx: 1,
        yy: 2,
    };
    let thing2 = Thing {
        xx: 3,
        yy: 4,
    };
    let things = [thing1, thing2];
    
    assert_eq(things[0].xx, xx);
    assert_eq(things[1].yy, yy);

    return things;
}
```

this one works as well^

erhant:
Have to admit that by "work" I only mean it compiles & a proof is generated and verified, not sure if there are some soundness on the backend side errors due to the way this is implemented 🤔

mimoo:
So Bool will work because a Bool is essentially a Field underneath it. Your second example should not work properly because it should allocate only 2 fields when it should actually allocate 4 fields (unless I'm missing something). This should be visible with a debug output (altho I think we don't have labels for r1cs, but the debug output of kimchi might show it better)

erhant:
> So Bool will work because a Bool is essentially a Field underneath it.

ah indeed!

> Can you post the r1cs debut output just for completeness as a comment here :D?

I dont know if I was dreaming yesterday (it was kinda late), but today I got the error:

```sh
Error:   × Looks like something went wrong in constraint-generation
    ╭─[examples/types_array_output.no:14:1]
 14 │     };
 15 │     let things = [thing1, thing2];
    ·                  ────────┬───────
    ·                          ╰── here
 16 │     
    ╰────
  help: the public output cannot contain constants
  ```
  
  After changing the circuit as below:
  
  ```rs
  struct Thing {
    xx: Field,
    yy: Field,
}

fn main(pub xx: Field, pub yy: Field) -> [Thing; 2] {
    let thing1 = Thing {
        xx: xx,
        yy: 1 + yy,
    };
    let thing2 = Thing {
        xx: 1 + xx,
        yy: yy,
    };
    let things = [thing1, thing2];
    
    assert_eq(things[0].xx, xx);
    assert_eq(things[1].yy, yy);

    return things;
}
  ```

the resulting R1CS is:

```
@ noname.0.7.0

v_3 == (v_3) * (1)
v_4 == (v_4) * (1)
v_3 == (v_1) * (1)
v_4 + 1 == (v_2) * (1)
```

Indeed it is not sized properly, and a test shows that only the fields of `thing1` are included in the output. I will try to tackle it 🙏🏻

**Comment by erhant** [comment]
Indeed recursing into items was a bit too much, I've added a check to only allow `Field` and `Bool` for now.

**Comment by mimoo** [comment]
there's also a `sizeof` or `size_of` function somewhere, do you think you could just use that to make it work with everything?

File: examples/fixture/asm/r1cs/public_output_array.asm [hunk 2]
```diff
@@ -0,0 +1,6 @@
+@ noname.0.7.0
+
+2 == (v_3 + v_4) * (1)
+v_5 == (v_3 + v_4) * (v_3)
+v_3 + v_4 + 6 == (v_1) * (1)
+v_5 == (v_2) * (1)
```
mimoo, line 6: ^ this last constraint seems really useless, we should optimize that, I'll create an issue :)

erhant:
> there's also a `sizeof` or `size_of` function somewhere, do you think you could just use that to make it work with everything?

yea that seems to solve things, I simply get the size of the returned type and add that many public outputs to the circuit.

However, there is still the issue of order of the public outputs during tests, it appears to be randomly shuffled (?) everytime, this is not apparent in the previous tests because all of them have at most 1 public output. I will write a separate test for this real quick.

erhant:
> However, there is still the issue of order of the public outputs during tests, it appears to be randomly shuffled (?) everytime, I will write a separate test for this real quick.

The issue seems to be about Kimchi backend, not R1CS. I further found the reason why, its this here: <https://github.com/zksecurity/noname/blob/main/src/backends/kimchi/mod.rs#L428>

Iterating the hashmap does not guarantee order, so this here fails.

erhant:
I made it use `BTreeMap` instead, which guarantees order; I made the ordering respect the `index` of the cell only, I think that is alright?

src/backends/kimchi/mod.rs
```diff
@@ -243,12 +243,18 @@ impl KimchiVesta {
...
     index: usize,
     pub span: Span,
 }
 
+impl PartialOrd for KimchiCellVar {
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        Some(self.index.cmp(&other.index))
+    }
+}
+
```
mimoo, line 257: that's great! but can you use this approach instead? https://github.com/zksecurity/noname/pull/108

I think it would look like:

```diff
#[educe(PartialCmp)]
pub struct Ident {
    index: usize,
+    #[educe(PartialCmp(ignore))]
    pub span: Span,
}
```

examples/fixture/asm/r1cs/public_output_array.asm
katat: Do you mean the second constraint should be instead 
`v_2 = (v_3 + v_4) * (v_3)`?

`v_1` and `v_2` are the outputs and they have to be constrained

mimoo (approved):
thanks so much for the PR :) !!

examples/fixture/asm/r1cs/public_output_array.asm
mimoo: yeah! `v_2 == (v_3 + v_4) * v_3` would have been better. But I don't think we should look to create such optimization, because an optimization pass that would do inlining would easily inline the last constraint

PR #136, opened 2024-06-23
Refactor with ComputedExpr

katat:
This is a prerequisite PR for the upcoming support of const generic feature #120


book/src/rfc/const_generic.md
```diff
@@ -0,0 +1,265 @@
+## Summary
+Enable const generic to improve modularization and reusability of the code.
```
mimoo, line 2: maybe a good idea to precise that this is const generics for functions, and not for types

book/src/rfc/const_generic.md
```diff
@@ -0,0 +1,265 @@
...
+
+## Motivation
+In the current implementation, array initialization requires explicitly stating all element values, which becomes impractical for large arrays. Furthermore, the lack of support for constant variables as array sizes limits the ability to write functions that generically handle arrays of varying sizes. This also necessitates hardcoding loop ranges, reducing flexibility.
```
mimoo, line 5: not-really-important comment: it's always a good idea to add example. For example, here you could add an example of what array initialization you're talking about. I think also in general it'd be nice to have the RFC be its own PR so that people can comment/improve on it without messing with the implementation PR. But let's keep this way this time :)

book/src/rfc/const_generic.md
```diff
@@ -0,0 +1,265 @@
...
+
+Enabling const generic will allow define array sizes with constant variables. This would enhance function reusability and significantly improve modularization.
+
+## Detailed design
+
```
mimoo, line 10: good to have intro text in each section/subsection, rather than empty text

book/src/rfc/const_generic.md
```diff
@@ -0,0 +1,265 @@
...
+### Applying a const as a argument
+Use an existing constant to set array size directly in function parameters:
+
+```rust
+const cst = 300;
+fn const_generic(yy: Field) -> [Field; cst] {
+    let xx = [1; cst];
+    return xx;
+}
+```
```
mimoo, line 33: so this wasn't working before?

mimoo: (btw this is not a const generic, this is just a constant, so this RFC is more general than just const generic)

book/src/rfc/const_generic.md
```diff
@@ -0,0 +1,265 @@
...
+
+### Builtin functions
+
+Allow builtin functions to be defined with const arguments:
```
mimoo, line 37: maybe rephrase: The function's signature is determined by a constant value that you can pass as argument"

book/src/rfc/const_generic.md
```diff
@@ -0,0 +1,265 @@
...
+
+```rust
+use std::to_bits;
+use std::from_bits;
+
+const num_bits = 8;
+
+fn main() {
+    let val1 = 101;
+    let bits = to_bits(num_bits, val); // bits is array ComputedExpr with len
```
mimoo, line 47: at this point `ComputedExpr` hasn't been defined. It's good to define terms right when we use them for the first time

mimoo: also here you mean "with length `num_bits` no?

book/src/rfc/const_generic.md
```diff
@@ -0,0 +1,265 @@
...
+    let val2 = from_bits(bits); // from_bits builtin impl knows the size of the array
```
mimoo, line 48: don't we need to pass `num_bits` as argument to make it work here as well? I guess builtins can be generic on the length of their argument without using const generics

mimoo: so maybe that RFC should be called something like "Supporting variable-length arguments and returned values in functions"

mimoo: I think it would still be more secure to pass a `const num_bits` argument to enforce the intent of the user. The same way [`memcpy_s`](https://en.cppreference.com/w/c/string/byte/memcpy) was invented because of the bugs people wrote using `memcpy`

book/src/rfc/const_generic.md
```diff
@@ -0,0 +1,265 @@
...
+    assert_eq!(val1, val2);
```
mimoo, line 49: another example of generic function that's generic thanks to builtins!

book/src/rfc/const_generic.md
```diff
@@ -0,0 +1,265 @@
...
+}
+```
+
+The `to_bits` builtin function accepts a const variable to determine the size of the return array.
+
+```rust
+fn to_bits(const num_bits: Field, val: Field) -> [Field; num_bits]
```
mimoo, line 56: ah right, I guess we still need a const generic syntax for builtins so that we can parse their signature for the type checker

book/src/rfc/const_generic.md
```diff
@@ -0,0 +1,265 @@
...
+```
+
+The from_bits builtin function can determine the number of bits to convert based on the information that the `ComputedExprKind::Array` embeds in the `bits` variable. We will explain on `ComputedExpr` at the end to show how to improve the structure of the inputs and outputs of these builtin functions.
```
mimoo, line 59:
```diff
- The from_bits ...
+ The `from_bits` ...
```

mimoo: first time `ComputedExprKind` is used, define here

mimoo: > We will explain on `ComputedExpr`

->

> We will explain `ComputedExpr`

or

> We will explain how `ComputedExpr` works

book/src/rfc/const_generic.md
```diff
@@ -0,0 +1,265 @@
...
+
+## Type checking const variable
+
+The existing type checker is designed to do local type checks. For the cases of const variable as array size, the type checker has to trace the values globally. This is out of the scope for the design of the current type checker.
```
mimoo, line 63: why do you say that it is out of scope, if you provide a solution in the next paragraph?

mimoo: also, we should write here what the typechecker can do with local information, it can still check that symbolic typing makes sense. Maybe give an example of something that should not work:

```rust
fn thing(const len: Field) -> [Field; len + 1] {
  return [0; len];
}
```

book/src/rfc/const_generic.md
```diff
@@ -0,0 +1,265 @@
...
+
+Instead of checking the const variable in the type checker, we can check it in the circuit generation phase.
```
mimoo, line 65: we never properly defined what a "const variable" was in this document BTW. Would be good to add something to the intro. Perhaps "`const` is a keyword that can be used in different places. For example, when used in front of an argument name, in a function signature, it dictates that the function must be called with a value decided at compile time (for example, with a literal, or a global constant)"

book/src/rfc/const_generic.md
```diff
@@ -0,0 +1,265 @@
...
+
+The from_bits builtin function can determine the number of bits to convert based on the information that the `ComputedExprKind::Array` embeds in the `bits` variable. We will explain on `ComputedExpr` at the end to show how to improve the structure of the inputs and outputs of these builtin functions.
+
+## Type checking const variable
+
+The existing type checker is designed to do local type checks. For the cases of const variable as array size, the type checker has to trace the values globally. This is out of the scope for the design of the current type checker.
+
+Instead of checking the const variable in the type checker, we can check it in the circuit generation phase. 
+
+### Review how VarOrRef works
```
mimoo, line 67:
```
`VarOrRef`
```

mimoo: maybe add some intro to explain what we're going to read about here: we're going to read about how noname currently works and support mutable accesses

mimoo: also would be good to have an explanation here on why this is relevant in the conversation on const generic

book/src/rfc/const_generic.md
```diff
@@ -0,0 +1,265 @@
...
+```rust
+houses[1].rooms[2].size
+```
+
+The variable `houses` holds all the `cvars` to represent the underlying low level variables. If `houses` is a mutable variable, it will be a reference, which records the following instead of holding the actual `cvars` array:
+ - the variable name `houses` in local scope
+ - start index of the `cvars` to represent the current access path
+ - the length of the `cvars` to represent size of the element targeted by the current access path
+
+So when it wants to update the elements inside the `houses`, it will need to reference the start index and the length of the `cvars` of the elements, so that it can narrow down the scope of the `cvars` to be updated.
```
mimoo, line 81: ah OK I think I understand why I initially did this, because when you do 

```rust
houses[1].rooms[2] = 5;
```

you end up parsing expression after expression recursively, and as you go from left to right you lose information. It sorts of look like this AST:

```
Assignment(ArrayAccess(FieldAccess(ArrayAccess("houses", 1), "rooms"), 2), 5)
```

and so you want to be able to pinpoint exactly what exactly is being assigned something (so that it can be reflected later in the scope where `houses` exist)

book/src/rfc/const_generic.md
```diff
@@ -0,0 +1,265 @@
...
+        }
+    }
+}
+```
+
+This allows to find the range of the `cvars` belonging to the element targeted by the current access path under the local variable behind the `var_name`, which in this case is the `houses`. 
+
+But it doesn't keep track of the structure of the data. For example, when we are accessing `houses[1].rooms[2]`, it should be able to check if the access is within the array bounds. Although the current design can check if it is out of bound by checking access index against the total number of `cvars` for `houses`, it can't label which element access is out of bound as it could be either out of bound of `houses[1]` or the nested access `rooms[2]`.
+
+### Introduce ComputedExpr
```
mimoo, line 132:
```
`ComputedExpr`
```

book/src/rfc/const_generic.md
```diff
@@ -0,0 +1,265 @@
...
+
+With the const variable as array size, we need to propagate its value during circuit generation to check if the access is within the bounds of the array. So we need a way to retain structure information of a variable that is currently represented by `VarOrRef`.
```
mimoo, line 134: not sure I understand why, if it's not in the bounds of the array you can see that pretty quickly as `some_array.get(some_idx)` will return `None`

book/src/rfc/const_generic.md
```diff
@@ -0,0 +1,265 @@
...
+
+We can add a new layer to the `VarOrRef` to keep track of the structure of the data. It is called `ComputedExpr`, as it is computed from `compute_expr` function, which resolves and passing around the values (`cvars`) via the `Expr`. The key difference from `VarOrRef` is `ComputedExpr` retains the structure of the underlying variables.
```
mimoo, line 136: I'm wondering if this is the kind of things that should prompt us to create an IR...

book/src/rfc/const_generic.md
```diff
@@ -0,0 +1,265 @@
...
+
+`ComputedExpr` enables the field / array access to determine the structure of the value with actual size of the targeted access. For example, when we are accessing `houses[1].rooms[2]`, it can check if the access is within the bounds of `houses` and `rooms`, by checking the structure and size of a targeted `ComputedExpr`.
+
+In a sense, `ComputedExpr` can be seen as an resolved `Expr` with const variables replaced with actual values.
```
mimoo, line 140: but so far that doesn't help explain how this is useful for const generics

book/src/rfc/const_generic.md
```diff
@@ -0,0 +1,265 @@
...
+
+With the const variable as array size, we need to propagate its value during circuit generation to check if the access is within the bounds of the array. So we need a way to retain structure information of a variable that is currently represented by `VarOrRef`.
+
+We can add a new layer to the `VarOrRef` to keep track of the structure of the data. It is called `ComputedExpr`, as it is computed from `compute_expr` function, which resolves and passing around the values (`cvars`) via the `Expr`. The key difference from `VarOrRef` is `ComputedExpr` retains the structure of the underlying variables.
+
+`ComputedExpr` enables the field / array access to determine the structure of the value with actual size of the targeted access. For example, when we are accessing `houses[1].rooms[2]`, it can check if the access is within the bounds of `houses` and `rooms`, by checking the structure and size of a targeted `ComputedExpr`.
+
+In a sense, `ComputedExpr` can be seen as an resolved `Expr` with const variables replaced with actual values. 
+
+### Types of ComputedExpr
```
mimoo, line 142:
```
`ComputedExpr`
```

book/src/rfc/const_generic.md
```diff
@@ -0,0 +1,265 @@
...
+
```
mimoo, line 143: this section needs some intro text

book/src/rfc/const_generic.md
```diff
@@ -0,0 +1,265 @@
...
+    Bool(Var<F, V>),
+    Field(Var<F, V>),
+    /// Access to a variable in the scope.
+    Access(Access<F, V>),
+    /// Represents the results of a builtin function call.
+    /// Because we don't know the exact type of the result, we store it as a Var.
+    /// We may deprecate this once it is able to type check the builtin functions,
+    /// so that the result can be inferred.
+    FnCallResult(Var<F, V>),
+}
```
mimoo, line 163: probably needs the `ComputedExpr` struct itself too

book/src/rfc/const_generic.md
```diff
@@ -0,0 +1,265 @@
...
+```
+
+### Assignment using ComputedExpr
+
+Instead of using `VarOrRef` to narrow down and do the updating of `cvars`, we  use the `ComputedExpr` to represent the access path, which is iteratively determined when stepping through the computations of `Expr::FieldAccess` and `Expr::ArrayAccess`.
+
+When it comes to the computation for `Expr::Assignment`, it can use the access path embeded in a `ComputedExprKind::Access` to locate a nested `ComputedExpr` to swap with a new one. Then it saves the `VarInfo` with same name `houses` as local variable but holding the updated tree of `ComputedExpr` that represents a new set of `cvars`.
```
mimoo, line 170: embedded*

book/src/rfc/const_generic.md
```diff
@@ -0,0 +1,265 @@
...
+        let zero_mul = compiler.backend.mul(bit, compiler.backend.sub(bit, 1));
+        compiler.backend.assert_eq_const(zero_mul, 0, span);
+
+        // lc += bits[i] * e2;
+        lc = compiler.backend.add(lc, compiler.backend.mul(bit, e2));
+
+        e2 = e2 + e2;
+    }
+
+    compiler.backend.assert_var!(value, lc, span);
```
mimoo, line 226: without the `!` no?

book/src/rfc/const_generic.md
katat: > so this wasn't working before?

It didn't support array init with default value by size `let xx = [1; cst];`
```

```

book/src/rfc/const_generic.md
katat: Exactly!

book/src/rfc/const_generic.md
```diff
@@ -0,0 +1,265 @@
...
+}
+```
+
+### Assignment using ComputedExpr
+
+Instead of using `VarOrRef` to narrow down and do the updating of `cvars`, we  use the `ComputedExpr` to represent the access path, which is iteratively determined when stepping through the computations of `Expr::FieldAccess` and `Expr::ArrayAccess`.
+
+When it comes to the computation for `Expr::Assignment`, it can use the access path embeded in a `ComputedExprKind::Access` to locate a nested `ComputedExpr` to swap with a new one. Then it saves the `VarInfo` with same name `houses` as local variable but holding the updated tree of `ComputedExpr` that represents a new set of `cvars`.
+
+### Use ComputedExpr in builtin functions
```
mimoo, line 172: so so far:

* I think I understand the usecases this unlock
* I'm not sure why `VarOfRef` is relevant to the explanation, but I understand what it does (maybe the RFC should mention that the main functions in the parser (`compile_stmt` and `compile_expr`) returns a `VarOrRef`
* I don't think the RFC spends enough time explaining why const generics can't work as is currently. IMO it should explain why it doesn't work in two different phases of the compilation: type checker and circuit writer
* this RFC is mostly on functions that take const args, but it doesn't really spell that

book/src/rfc/const_generic.md
mimoo: also at this point I don't understand why a `ComputedExpr` has so many tags

book/src/rfc/const_generic.md
katat: I probably need to rephrase this one also. But I meant during `ExprKind::Array` it was using `size_of` for a type to determine the size of the array. With the support of const variable length for array, the actual value of the const variable needs to be propagated to `ExprKind::Array` computation.

book/src/rfc/const_generic.md
katat: I shifted these examples to the bottom.

book/src/rfc/const_generic.md
katat: added it to the glossary section

book/src/rfc/const_generic.md
katat: I replaced this section with `Resolving const variables`, aiming to simplify the explanation by focusing the current use of the `size_of` and the needs to retain the structure from bottom up.

book/src/rfc/const_generic.md
katat: It is because it can't rely on the `size_of` to do the check anymore for arrays with variable length. It was using `size_of` to determine the size of `some_array`.

I detailed the explain in the `Resolving const variables`.

book/src/rfc/const_generic.md
katat: I am sure `compute_expr` will be the focus when doing integration with circify ;)

book/src/rfc/const_generic.md
katat: For the array size case, it is about resolving the values without having to rely on `size_of`. To do so, it requires retaining the structure information from the bottom up.

book/src/rfc/const_generic.md
katat: you really got a great private parser!!

book/src/rfc/const_generic_fn.md
```diff
@@ -0,0 +1,224 @@
...
+
+### Glossary
+- const: a keyword that can be used in different places. For example, when used in front of an argument name, in a function signature, it dictates that the function must be called with a value decided at compile time (for example, with a literal, or a global constant)"
+
+### Resolving const variables
+When the size of an array declaration is a literal number, the size can be resolved by calling the `size_of` on the array type. Then the `ExprKind::ArrayAccess` can check the bounds of the access by referencing the number returned from `size_of` call on an array type.
+
+When it is a constant variable as array size, we can't rely on the `size_of` to determine the size of the array directly through the type anymore, as the type checker can't resolve these values under its current design. For example, the type checker can't determine wether the length of the array matches with the return type:
```
mimoo, line 20: I feel like you're talking about initializing the array, but the example shows an issue of type checking. Shouldn't you write:

```diff
fn thing(const len: Field) -> [Field; len] {
-  return [0; 3];
+  return [0; len];
}
```

as example instead?

book/src/rfc/const_generic_fn.md
```diff
@@ -0,0 +1,224 @@
...
+
+```rust
+fn thing(const len: Field) -> [Field; len] {
+  return [0; 3];
+}
+```
+
+That is because the type checker doesn't know the actual value of `len`, which needs to be propagated when computing the `Expr` during circuit generation.
+
+This RFC proposal is to defer the resolution of the actual values for the variable sizes of arrays to the circuit generation phase. The resolution can be done via propagating the actual sizes from the bottom up. This means to store the resolved values in a structured way, so they can be accessible in the computations of `compute_expr`, which is to compute the structured expressions.
```
mimoo, line 30: `compute_expr` is not a type checker function, this is why I'm confused by this section

mimoo: IMO we should split the problem in two parts:

1. what does type checking needs to do? And what prevents it from doing its job with the current impl
2. what does circuit synthesizer needs to do? And what prevents it from doing its job with the current impl

I'm not sure I understand the answer to both question ATM.

mimoo: another question I have is, why not solve this by using two different `size_of` functions for type checking and circuit synthesizing? Or perhaps augment that function with a `ctx: Option<_>` argument which is `None` at type checking, but contains the actual instantiation at circuit synthesizing. Something like this perhaps:

```rust
#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash)]
pub enum TyKind {
    // TRUNCATED...
    /// An array of a fixed size.
    Array(Box<TyKind>, u32),
    /// An array of a generic size.
    GenericArray(Box<TyKind>, String),
    // TRUNCATED...
}

// we assume that Mul and Add are implemented for this type
enum TySize {
    Concrete(usize),
    Generic(String),
    Mul(TySize, TySize),
    Add(TySize, TySize),
}

impl<B: Backend> TypeChecker<B> {
    pub(crate) fn size_of_concrete(&self, ctx: HashMap<String, usize>, typ: &TyKind) -> usize {
        match self.size_of(Some(ctx), typ) {
            TySize::Concrete(size) => size,
            _ => panic!("this should not happen"),
        }
    }

    pub(crate) fn size_of(&self, ctx: Option<HashMap<String, usize>>, typ: &TyKind) -> TySize {
            match typ {
                // TRUNCATED...
                TyKind::GenericArray(typ, len_var) => {
                    if let Some(ctx) = ctx {
                        let size_of_typ = match self.size_of(ctx, typ) {
                            TySize::Concrete(size) => size,
                            _ => panic!("this should not happen"),
                        };
                        TySize::Concrete(ctx[len_var] * size_of_typ)
                    } else {
                        TySize::Mul(
                            self.size_of(None, typ),
                            TySize::Generic(len_var.clone()),
                        )
                    }
                }
            }
        }
```

mimoo (commented):
more comments, I think we should probably have a sync conversation to resolve these :o

book/src/rfc/const_generic_fn.md
```diff
@@ -0,0 +1,224 @@
...
+    compiler: &mut CircuitWriter<B>,
+    vars: &[VarInfo<B::Field, B::Var>],
+    span: Span,
+) -> Result<Option<ComputedExpr<B::Field, B::Var>>> {
+    // should be two input vars
+    assert_eq!(vars.len(), 2);
+
+    // first var is the number of bits
+    let num_var = &vars[0];
+    assert!(matches!(num_var.expr.kind, ComputedExprKind::Field));
```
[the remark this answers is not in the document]
katat: Yeah, these checks should be already covered by the type checker. They are probably unnecessary here.

On the related issue, I think we have to check the const generic var is consistent with the actual array size resulted from the function call:

```rust
fn to_bits(const num_bits: Field, val: Field) -> [Field; num_bits]
```

Because the type checker doesn't know the actual value of `num_bits`, we would need to check the provided const generic argument against the size of the resulted array.

book/src/rfc/const_generic_fn.md
katat: In terms of the use of `ctx`, I think here is a scenario it wouldn't work:

```rust
fn gen_arr(const size) -> [field; size] {
	return [0; size]; 
	// the FnEnv only store local context
	// so it either needs to:
	// 1. let VarOrRef carries the size value
	// 2. let FnEnv store additional meta info for VarOrRef, but the problem is FnEnv is designed to be scoped. It will need to propagate the meta info from this local scope to outer scopes, which is complicated and doesn't align with the purpose of FnEnv.
}

fn compare_arr(const size, a: [field; size], b: [field; size]) -> bool {
	...
}

let arr1 = gen_arr(10); // arr1 should carry the meta info for size

let arr2 = [0; 9];

// so the meta info can be used to check the arguments
assert_eq(compare_arr(arr1, arr2), true); 
```

I think we need to let the `VarOrRef` to carry the additional info, as these variables are passing between different scopes, among which it seems to be difficult to trace values directly via `FnEnv` which is scoped. But to achieve that with `VarOrRef`, `VarOrRef` has to be structured like `ComputedExpr`.

So I left this approach for now, and am investigating the other approaches.

katat: yeah, but it might work if we require the front-end syntax to enforce explicit declaration like
```rust
let arr1: [field; 10] = gen_arr(10);
```

I am not sure if this can complicate the dev experience. This requires the coders to do the calculation themselves, and this can be very cumbersome, especially const value returned from a generic function involves complicated calculations, such as 

```rust
fn gen_arr(const size) -> [field; size * 123] {
```

mimoo: my take is that monomorphization is probably the way, and if we need to force the dev to write the types to get that information we should do that (we can always implement inference later if we want it)

katat:
implemented by another approach: https://github.com/zksecurity/noname/pull/147

PR #142, opened 2024-07-08
RFC: generic-sized arrays

katat:
This is an alternative approach to #136, based on the _monomorphization_ concept. 
It aims to have broader support for const generic.


book/src/rfc/rfc-0-generic-parameters.md
```diff
@@ -0,0 +1,413 @@
...
+- Typecheck the inferred Array type (or anything can't be checked in TAST should be checked in MAST)
+- Return MAST for circuit synthesizer
+
+With the type information collected from the TAST phase, the MAST phase can infer the generic values from the observed arguments by running through the `check_block` function, which is used to be done in TAST phase. The `check_block` is context aware, and will recursively walk through the main function body. We can devise an algorithm to incorporate with `check_block` to infer the generic values and store in a certain way for later use.
+
+
+### Inferring algorithm
+The algorithm will need to handle the following inference categories:
+- Inferencing from constant argument
+- Inferencing from array arugment
```
mimoo, line 239: argument*

book/src/rfc/rfc-0-generic-parameters.md
```diff
@@ -0,0 +1,413 @@
...
+    // type of house1 is Array(Room, 3)
+    // type of house2 is Array(Room, 1)
+    let rooms = all_rooms(house1, house2);
+    ...
+}
+```
+
+The current `compute_type` function can remain as it is. It stores the computed concrete types, such as `Array` or `Custom` as shown in the examples above, instead of the generic types. It creates a mapping between the expression node from the AST and the type `TypeKind`.
+
+### Circuit Synthesizer
```
mimoo, line 387: so my understanding was that there would be no change to the circuit synthesizer thanks to monomorphization

mimoo (commented):
looks like this is the right direction! Looking forward to the PR :)

book/src/rfc/rfc-0-generic-parameters.md
```diff
@@ -0,0 +1,413 @@
...
+
+// the return type is determined by the generic arguments
+// so `init_arr` can be used to create arrays with different sizes
+fn init_arr<N, M>(n: N, m: M) -> [Field; N * 2 + M] {
+    let arr = [0; n * c_mul + m];
+    return arr;
+}
+
+
+fn main() -> [Field; 3 * c_mul + 1] {
```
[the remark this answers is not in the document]
katat: Agreed. I will simplify this example without involving the constant var.

I think #143 requires changes to:
- update AST to support expression as the size of the array type.
- update `compute_type` in the type checker to resolve the value from the expression before resolving it to `Array(TyKind; Size)`.

book/src/rfc/rfc-0-generic-parameters.md
```diff
@@ -75,6 +75,117 @@ fn main() -> [Room; 4] {
...
+Given the following function signatures for builtin functions:
+```rust
+fn from_bits(const N: Field, val: Field) -> [Field; N]
+fn from_bits(bits: [Bool; N]) -> Field
+```
+
+Calling the builtin functions in native code:
+```rust
+use std::to_bits;
+use std::from_bits;
```
mimoo, line 88: syntax should be:

```rust
use std::bits; // import
bits::to_bits(_); // usage
```

(this is to mimic golang)

book/src/rfc/rfc-0-generic-parameters.md
```diff
@@ -75,6 +75,117 @@ fn main() -> [Room; 4] {
 }
 ```
 
+## Builtin Examples
+Given the following function signatures for builtin functions:
+```rust
+fn from_bits(const N: Field, val: Field) -> [Field; N]
```
mimoo, line 81: to_bits*

book/src/rfc/rfc-0-generic-parameters.md
mimoo: BTW I think we should have a syntax for builtins in general, so that `std::bits` can mix both builtins and native code. For example, the `bits` module could look like this:

```rust
fn not(bit: Bool) -> Bool {
  return 1 - bit;
}

builtin fn to_bits(const N: Field, val: Field) -> [Field; N];
```

with a `builtin` keyword?

katat: lgtm!

book/src/rfc/rfc-0-generic-parameters.md
```diff
@@ -519,6 +519,35 @@ fn main() {
...
+fn main() {
+    let arr1 = [1, 2, 3, 4, 5];
+    let arr2 = [6, 7, 8, 9];
+
+    let last1 = last(arr1); // instantiate `last` function with N = 5
+    let last2 = last(arr2); // instantiate `last` function with N = 4
+}
+```
+
+The monomorphized body of the function call for `last(arr1)` is `return arr[5 - 1]`, while the one for `last(arr2)` is `return arr[4 - 1]`. Therefore, we can't have a single expression node representing both `arr[5 - 1]` and `arr[4 - 1]` expression nodes. These functions should be instantiated with new ASTs, which are monomorphized from the original ASTs. They will be regenerated with the generic parameters being resolved with concrete values.
```
mimoo, line 504: ... and the two calls to the `last()` functions should be made to the two different monomorphized instances

book/src/rfc/rfc-0-generic-parameters.md
```diff
@@ -519,6 +519,35 @@ fn main() {
...
+    let arr2 = [6, 7, 8, 9];
+
+    let last1 = last(arr1); // instantiate `last` function with N = 5
+    let last2 = last(arr2); // instantiate `last` function with N = 4
+}
+```
+
+The monomorphized body of the function call for `last(arr1)` is `return arr[5 - 1]`, while the one for `last(arr2)` is `return arr[4 - 1]`. Therefore, we can't have a single expression node representing both `arr[5 - 1]` and `arr[4 - 1]` expression nodes. These functions should be instantiated with new ASTs, which are monomorphized from the original ASTs. They will be regenerated with the generic parameters being resolved with concrete values. 
+
+To ensure no conflicts in the node IDs being regenerated for these instantiated functions, the AST for the main function as an entry point will be regenerated. The monomorphized AST preserves the span information to point to the noname source code for the existing debugging feature.
```
mimoo, line 506: BTW maybe a good thing to recap what is a node ID and what is stored under a node ID.

mimoo: to my understanding, the parser assigns a node_id to every node in the AST it creates, then the type checker stores some information on an AST node via its node id:

```rust
pub struct TypeChecker<B: Backend> {
  node_types: HashMap<usize, TyKind>,
```

mimoo: knowing this, can't we preserve the node id of the two calls to `last()` but fix `node_types` so that it returns the correct `Type`? If the synthesizer/circuit-writer tries to access its type later, and it's still `GenericArray` then it should panic.

book/src/rfc/rfc-0-generic-parameters.md
```diff
@@ -519,6 +519,35 @@ fn main() {
...
+
+Same as before, these instantiated functions can be pointed by the expression nodes `ExprKind::FnCall`. With the support of generic parameters, we need to change the way of loading the function AST, as the current fully qualified name pattern doesn't contain the information to differentiate the instantiated functions with different generic values.
+
+The new string pattern to store the monomorphized function AST can be:
+`fn_full_qualified_name#generic1=value1#generic2=value2`
```
mimoo, line 511: 👍

mimoo: so can you add a blurb right here on how the monormophization really works? My idea now is that monomorphization walks through the AST and:

1. propagate types in the same way the type checker was doing but replaces any generic type with a concrete type + folds constants + performs some type checking when it converts a generic type into a concrete type
2. whenever it encounters a generic function call it copies its AST into a new function `new_name = monomorphize_name(function_name, generics_and_values)` stored in the AST, then walks through its AST (modification happening due to 1 would apply on the AST of `new_name`. 
3. at the end it deletes every generic functions from the AST

the important thing is that in 1. when a generic type is converted into a concrete type, if we are inside of a generic function then we need to assign the node a new node ID

mimoo: so just to drive the point I was making last time, I think with that change we could have that the monomorphization pass only mutates the TAST, making it a simple wrapper at the end: `Mast<B: Backend>(TypeChecker<B>)`

It would only ever mutate the TAST to:

1. add new functions in `functions: HashMap<FullyQualified, FnInfo<B>>,`
2. convert generic types into concrete types in `node_types: HashMap<usize, TyKind>,`
3. assign a new node ID to instantiated/monomorphized functions nodes that are generic types and register them in `node_types: HashMap<usize, TyKind>,`

mimoo:
changed the name to "generic-sized arrays" as this is what we are really targeting afterall!

book/src/rfc/rfc-0-generic-parameters.md
```diff
@@ -0,0 +1,413 @@
...
+```rust
+// N can be inferred from the array size of argument `arr`
+fn last<N>(arr: [Field; N]) -> Field {
+    // use generic parameter N to form dynamic expressions in the function scope
+    return arr[N - 1];
+}
+
+fn main() -> Field {
+    let arr = [1, 2, 3, 4, 5];
+    return last(arr);
```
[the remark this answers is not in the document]
katat: I think as along as the symbolic size doesn't involve operations, we should be able to extract the generic values in a recursive fashion.

book/src/rfc/rfc-0-generic-parameters.md
```diff
@@ -0,0 +1,413 @@
...
+        rooms: [Room { ... }],
+    };
+    // it will type check the inferred type is [Room; 4]
+    return all_rooms(house1, house2);
+}
+```
+
+
+## monomorphization
+The inference of the generic values can be done by observing the arguments passed to the function. Then it stores the inferred values in the relevant contexts for the following compiler pipeline to do type checking and circuit synthesizing. We call the process of inferring the generic values and type checking _Monomorphization_.
```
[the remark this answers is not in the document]
katat: maybe use the term "resolve"?

book/src/rfc/rfc-0-generic-parameters.md
katat: Added.

We also have to modify the `FnCall` node, otherwise it won't point to the monomorphized function ast.

book/src/rfc/rfc-0-generic-parameters.md
katat: Yes, we fix the node id in the implementation.

book/src/rfc/rfc-0-generic-parameters.md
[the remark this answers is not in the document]
mimoo: ^ I think my comment was before we introduced `const arg: Field` arguments :o

book/src/rfc/rfc-0-generic-parameters.md
```diff
@@ -0,0 +1,359 @@
...
+1. Parse the code into AST
+2. Convert the AST into NAST with naming resolution
+3. Convert the NAST into TAST with type metadata collection and type checking
+4. Circuit synthesizing TAST into an constraint system
+
+With generic parameters, the current TAST phase can't handle the type checking anymore, because the generic parameters are unknown. For example, it can't type check the array with symbolic size without resolving the values for the generic parameters.
+
+To extend the type checking to support generic parameters, we can add a MAST phase (Monomorphized AST), right after the TAST phase, to resolve the values of the generic parameters. The circuit synthesizer will rely on the MAST instead of the TAST to compile a circuit.
+
+The new pipeline will be: AST Parser -> NAST -> TAST -> MAST -> Circuit synthesizing
```
mimoo, line 103: Maybe we could use the diagram that I made

book/src/rfc/rfc-0-generic-parameters.md
```diff
@@ -0,0 +1,359 @@
...
+
+## Implementation
+To support the generic syntax as shown in the examples above, we need to make changes to the AST Parser support generic syntax. Furthermore, because the generic parameters can't be resolved at TAST phase, the some type checkings will be less strict and deferred to MAST phase.
+
+The newly added phase MAST will be responsible for resolving the generic values from the observed arguments. It includes type checking on the monomorphized types that are bypass in the TAST phase.
```
mimoo, line 108: I was reading this to try to understand exactly what we can’t typecheck before the MAST pass (and need to defer to the MAST pass). But it’s not entirely clear to me here. Maybe list exactly the checks that we need to do and can’t do at the TAST pass?

book/src/rfc/rfc-0-generic-parameters.md
```diff
@@ -0,0 +1,359 @@
...
+```rust
+// Then, here is a more challenging case that would require techniques like SMT to solve the unique values of the generic parameters.
+// Even LEN * 2 looks obvious to solve, solving it may need something like (computer algebra system) CAS in rust.
+// It is easy to evaluate the Symbolic values using the solved generic value. But the way around is difficult.
+fn last(arr: [Field; LEN * 2]) -> Field {
+    return arr[LEN - 1];
+}
+```
+
+In this RFC, we want to just enforce the syntax to be sufficient to support the simple cases. That is to disallow arithemtic operations among generic parameters for function arguments.
```
mimoo, line 267: Arithmetic*

book/src/rfc/rfc-0-generic-parameters.md
```diff
@@ -0,0 +1,359 @@
...
+// collect observed arguments
+//  at function callee scope
+//    for each argument
+//      if the argument involves generic parameters
+//        resolve the generic values from observed arguments
+//    for each statement
+//      compute the type with the resolved generic values
+//      type checks
+//  return type
+```
```
mimoo, line 282: Maybe a list of text instead of a codeblock of indented comments would be better :p

book/src/rfc/rfc-0-generic-parameters.md
```diff
@@ -0,0 +1,359 @@
...
+
+2. Along the way, it also propagates the AST nodes. When it is not part of a generic function, the node should remain the same. Otherwise, the node should be regenerated.
+
+3. Whenever it encounters a generic function call, it instantiates the function based on the arguments and store it as a new function with a monomorphized name, then walks through the instantiated function AST. The function call AST node will be modified with the monomorphized name while retaining the same node id and span.
+
+4. In the function instantiation process, all the AST nodes will be regenerated. This new AST will be stored under the monomorphized function name.
+
+5. After monomorphized a function, it should add the name of the original function to a list that records which function AST to delete at the end. We can't not delete the original function AST immediately, because it might be called at different places.
+
+6. In each function block scope, it should type check the return types, by comparing the propagated return type and the defined return type. All these types should be in concrete form without generic parameters involved.
```
mimoo, line 331: Thinking more about it I feel like we should be able to typecheck types symbolically and that should be enough… I think my understanding now is that we don’t really need to defer any type checking

book/src/rfc/rfc-0-generic-parameters.md
katat: added

book/src/rfc/rfc-0-generic-parameters.md
katat: added

book/src/rfc/rfc-0-generic-parameters.md
katat: See above the added list of cases where the type checks can't be done in tast. 
I think it is possible to type check symbolically within a local function scope, since the generic parameters in the scope can remain the same identifiers.
But when it comes to cases that are across function scopes, I am not sure how the equality can be checked as the generic parameters could mean differently in different function scope.

mimoo (approved):
let's merge this! great RFC!

PR #152, opened 2024-07-23
fix for duplicate declarations

### LINKED ISSUE #150: Can't use same for loop var
given:

```rust
fn main(pub public_input: Field, private_input: [Field; 3]) {
    let mut sum = 0;

    for ii in 0..3 {
        sum = sum + private_input[ii];
    
    }
    for ii in 0..3 {
        sum = sum + private_input[ii];
    }

    assert_eq(sum, public_input);
}
```

it throws error at the second for loop:
```
   ╭─[for_loop:7:1]
 7 │     }
 8 │     for ii in 0..3 {
   ·         ─┬
   ·          ╰── here
 9 │         sum = sum + private_input[ii];
   ╰────
  help: the variable `ii` is declared twice
```

The cause is the `TypedFnEnv`  use var name as key to store vars, but it doesn't remove the vars that are out of scope.

Prabhat1308:
fixes #150 .


src/type_checker/fn_env.rs
```diff
@@ -80,14 +75,12 @@ impl TypedFnEnv {
 
     /// Exits a scoped block.
     pub fn pop(&mut self) {
-        self.current_scope.checked_sub(1).expect("scope bug");
+        self.current_scope = self.current_scope.checked_sub(1).expect("scope bug");
```
mimoo, line 78: woop! did we have a bug here o_O?

Prabhat1308: Yes. It was not  assigning the result  after the subtraction resulting in abnormal values of `current_scope` .

examples/example.no
```diff
@@ -1,9 +1,16 @@
 use std::crypto;
 
-fn main(pub public_input: Field, private_input: [Field; 2]) {
-    let xx = private_input[0] + private_input[1];
-    assert_eq(xx, 2);
+fn main(pub public_input: Field, private_input: [Field; 3]) {
```
katat, line 3: please create a new test, instead of changing this existing one.

katat (commented):
This is awesome!
The only comment I have is to create a new test for this case instead of modifying the `example.no`.

examples/dup_var.no
```diff
@@ -0,0 +1,13 @@
+fn main(pub public_input: Field, private_input: [Field; 3]) {
```
katat, line 1: please add a test in the `tests/examples.rs` to point to this test code.

Prabhat1308: Added the test .

katat (approved):
Great, thanks!

PR #153, opened 2024-07-24
Generic syntax refactor for the ongoing PR for rfc0

katat:


examples/generic_array_access.no
```diff
@@ -1,5 +1,5 @@
-fn last<NN>(arr: [Field; NN]) -> Field {
-    return arr[NN - 1];
+fn last(arr: [Field; N]) -> Field {
+    return arr[N - 1];
```
mimoo, line 2: since we're yelling on one-letter variables, I think we should do the same on one-letter numeric generics, wdyt?

katat: Agreed. 
Actually it is already updated to allow one-letter numeric generic: https://github.com/zksecurity/noname/pull/153/files#diff-52652f71bed52ba289f8c2066c3f850de28f9c099e199f94f305b097212a915a

mimoo: no I meant we should forbid one letter numerics, to encourage real variable names. Like `Len` instead.

mimoo: but I don't know if it's a good idea :D I don't know... wdyt?

src/mast/mod.rs
```diff
@@ -951,20 +994,35 @@ impl<B: Backend> Mast<B> {
...
                     // store concrete array type for the argument name
                     let arr_mty = MTypeInfo::new(arr_type, *span, None);
-                    typed_fn_env.store_type(&sig_arg.name.value, &arr_mty)?;
+                    typed_fn_env.store_type(arg_name, &arr_mty)?;
                 }
                 _ => {
                     let typ = type_info.typ.as_ref().expect("expected a value");
                     let cst = type_info.constant;
+
+                    if is_generic_parameter(arg_name) {
```
mimoo, line 1012: maybe add a comment here:

> we expect `const TypeParam: Field` or `const normal_const: Field` or `normal_arg: SomeType`

although shouldn't that check be like in the parser? here we should just have a branch for that:

```rust
_ if is_generic_parameter(arg_name) && type_info.constant.is_some() => { /* ... */ }
_ => { /* ... */ }
```

src/mast/mod.rs
```diff
@@ -168,6 +171,54 @@ impl MastCtx {
...
+                    return Ok(());
+                }
+
+                Err(Error::new(
+                    "mast",
+                    ErrorKind::ConflictGenericValue(name, *v, value),
+                    span,
+                ))
+            }
+            Some(GenericValue::Unbound) => Ok(()),
```
mimoo, line 212: isn't this where you'd want to set the value?

src/mast/mod.rs
```diff
@@ -987,11 +1045,8 @@ impl<B: Backend> Mast<B> {
         // evaluate generic return types using inferred values
         let ret_ty = match &fn_sig.return_type {
```
mimoo, line 1046: this will need to be recursive also (as I pointed in the other PR), but unrelated to this PR

src/mast/mod.rs
```diff
@@ -1009,13 +1064,12 @@ impl<B: Backend> Mast<B> {
                 let sig_typed = FnSig {
                     name: sig.name.clone(),
                     kind: sig.kind.clone(),
-                    generics: HashSet::new(),
+                    generics,
```
mimoo, line 1067: this seem to have no effect due to not mutating generics

src/parser/types.rs
```diff
@@ -399,7 +413,11 @@ impl Ty {
                     // otherwise assume it is a GenericArray
                     // [Field; NN]
                     //         ^^^
-                    (TokenKind::Generic(name), TokenKind::RightBracket) => {
+                    (TokenKind::Identifier(name), TokenKind::RightBracket) => {
+                        if !is_generic_parameter(name) {
```
mimoo, line 417: this is valid code but will error here:

```rust
const len = 3;

fn thing([Field; len]) { /* ... */ }
```

src/parser/types.rs
```diff
@@ -419,7 +437,12 @@ impl Ty {
                                     ctx.error(ErrorKind::InvalidArraySize, siz_first.span)
                                 })?)
                             }
-                            TokenKind::Generic(name) => {
```
mimoo, line 422: personally I'm not a fan of this branch :D I would be happier if we didn't allow operating on generics in types (like in Rust) and maybe we can try to have this in a later PR. But this monomorphization change is trying to do too many things already IMO

src/parser/types.rs
```diff
@@ -507,7 +554,7 @@ impl Ty {
 //~
 //~ Backus–Naur Form (BNF) grammar:
 //~
-//~ fn_sig ::= ident ["<" ident "," ident ">"] "(" param { "," param } ")" [ return_val ]
+//~ fn_sig ::= ident "(" param { "," param } ")" [ return_val ]
 //~ return_val ::= "->" type
 //~ param ::= { "pub" } ident ":" type
```
mimoo, line 559: `"pub" | "const"` I guess :D but this is not up to date anyway

src/syntax.rs
```diff
@@ -40,23 +40,33 @@ pub fn is_identifier(s: &str) -> bool {
             .all(|c| (c.is_ascii_alphabetic() && c.is_lowercase()) || c.is_numeric() || c == '_')
 }
 
-// todo: add camel case check in is_type to exclude generic pattern from type's
 /// Returns true if the given string is generic parameter
 pub fn is_generic_parameter(s: &str) -> bool {
+    // should only be one uppercase letter
+    if s.len() != 1 {
+        return false;
+    }
```
mimoo, line 48: ah, maybe that is fine as well

src/parser/types.rs
```diff
@@ -516,10 +563,30 @@ impl FnSig {
...
 
-        let generics = FunctionDef::parse_generics(ctx, tokens)?;
-
         let arguments = FunctionDef::parse_args(ctx, tokens, &kind)?;
 
+        // extract generic parameters from arguments
+        let mut generics = GenericParameters::default();
+        for arg in &arguments {
+            match &arg.typ.kind {
+                TyKind::Field => {
```
mimoo, line 572: shouldn't we check that it is `const` as well? (e.g. `const N: Field`)

src/parser/types.rs
```diff
@@ -516,10 +563,30 @@ impl FnSig {
...
+                    // extract from const argument
+                    if is_generic_parameter(&arg.name.value) {
+                        generics.add(arg.name.value.to_string());
+                    }
+                }
+                TyKind::GenericArray(_, sym) => {
```
mimoo, line 578: doesn't take into account things like `[[Field; N]; 3]`

src/parser/types.rs
```diff
@@ -619,12 +686,21 @@ impl Default for FuncOrMethod {
...
 }
 
+#[derive(Debug, Default, Clone, Serialize, Deserialize)]
+pub struct GenericParameters(pub HashMap<String, GenericValue>);
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub enum GenericValue {
+    Bound(u32),
+    Unbound,
+}
```
mimoo, line 696: this type is the same as the Option type IMO

src/stdlib/mod.rs
```diff
@@ -134,6 +135,7 @@ fn assert_eq<B: Backend>(
 /// Asserts that a condition is true.
 fn assert<B: Backend>(
     compiler: &mut CircuitWriter<B>,
+    _generics: &GenericParameters,
```
mimoo, line 138: I'm wondering why we're doing this

src/mast/mod.rs
mimoo: I think you would have caught that if you had returned an error earlier if all the generics are not instantiated at this point

src/type_checker/checker.rs
```diff
@@ -641,6 +638,9 @@ impl<B: Backend> TypeChecker<B> {
         Ok(None)
     }
 
+    // todo: collect generic parameters and update FnInfo.generics
+    // - HashSet<Ident, ValueBound>
+    // - ValueBound: Enum { Bound(u32), Unbound }
```
mimoo, line 643: why?

src/type_checker/mod.rs
```diff
@@ -333,16 +333,20 @@ impl<B: Backend> TypeChecker<B> {
...
 
                     // store generic parameters as local vars in the fn_env
-                    for gen in &function.sig.generics {
+                    for name in function.sig.generics.names() {
                         typed_fn_env.store_type(
-                            gen.to_string(),
-                            // todo: let generic vars carry their span
-                            TypeInfo::new(TyKind::Generic(gen.to_string()), function.span),
+                            name.to_string(),
+                            TypeInfo::new_cst(TyKind::Field, function.span),
```
mimoo, line 339: `name.span`, the generics should be `Ident`

src/parser/types.rs
```diff
@@ -619,12 +686,21 @@ impl Default for FuncOrMethod {
     }
 }
 
+#[derive(Debug, Default, Clone, Serialize, Deserialize)]
+pub struct GenericParameters(pub HashMap<String, GenericValue>);
```
mimoo, line 690: I think it should be `HashMap<Ident, Option<u32>>`

src/type_checker/mod.rs
```diff
@@ -333,16 +333,20 @@ impl<B: Backend> TypeChecker<B> {
...
                         )?;
                     }
 
                     // store variables and their types in the fn_env
                     for arg in &function.sig.arguments {
+                        // skip const generic as they should be already stored
+                        if is_generic_parameter(&arg.name.value) {
+                            continue;
```
mimoo, line 347: maybe `assert!(function.sig.generics.names().contains(arg.name.value))`

src/mast/ast.rs
```diff
@@ -5,14 +5,10 @@ use super::MastCtx;
...
-        let node_id = ctx.next_node_id();
-        let span = self.span;
-        let kind = kind.clone();
-
         Expr {
-            node_id,
-            span,
-            kind,
+            node_id: ctx.next_node_id(),
+            kind: kind.clone(),
```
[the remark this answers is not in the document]
katat: It will also replace:
- BigInt (if it is a generic parameter)
- FnCall (the generic arguments will be replaced)

So anything that will need to be converted to concrete type will be replaced.

src/mast/mod.rs
```diff
@@ -168,6 +171,54 @@ impl MastCtx {
...
+impl Symbolic {
+    /// Evaluate symbolic size to an integer.
+    pub fn eval(&self, typed_fn_env: &MonomorphizedFnEnv) -> u32 {
+        match self {
+            Symbolic::Concrete(v) => *v,
+            Symbolic::Generic(g) => typed_fn_env.get_type_info(&g.value).unwrap().value.unwrap(),
+            Symbolic::Add(a, b) => a.eval(typed_fn_env) + b.eval(typed_fn_env),
+            Symbolic::Mul(a, b) => a.eval(typed_fn_env) * b.eval(typed_fn_env),
+        }
+    }
```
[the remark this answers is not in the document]
katat: This can be type checked in mast

src/mast/mod.rs
```diff
@@ -168,6 +171,54 @@ impl MastCtx {
...
+}
+
+impl GenericParameters {
```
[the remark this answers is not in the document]
katat: can you elaborate on this?

src/mast/mod.rs
```diff
@@ -168,6 +171,54 @@ impl MastCtx {
...
+        self.0.keys().cloned().collect()
+    }
+
+    /// Add an unbound generic parameter
+    pub fn add(&mut self, name: String) {
+        self.0.insert(name, GenericValue::Unbound);
+    }
+
+    /// Bind a generic parameter to a value
+    pub fn bind(&mut self, name: String, value: u32, span: Span) -> Result<()> {
```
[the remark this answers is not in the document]
katat: how about "assign", since it is assigning value to a parameter

src/lexer/mod.rs
```diff
@@ -248,14 +246,7 @@ impl Token {
...
                 } else if is_generic_parameter(&ident_or_number) {
-                    if ident_or_number.len() < 2 {
-                        return Err(ctx.error(
-                            ErrorKind::NoOneLetterVariable,
-                            Span::new(ctx.filename_id, ctx.offset, 1),
-                        ));
-                    }
-
-                    TokenKind::Generic(ident_or_number)
+                    TokenKind::Identifier(ident_or_number)
```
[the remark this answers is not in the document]
katat: the string rule for generic parameters is:
- all letters should be capitalized
- at least two letters

So the `is_identifier_or_type` can't pick up the generic parameters. It relies on the `is_generic_parameter` to check if it is generic-related at places where `TokenKind::Generic` and `TyKind::Generic` were used.

src/mast/mod.rs
katat: The check on the constant value is to make sure there is a constant value propagated.
Maybe this is unnecessary, as this should be checked at TAST.

src/mast/mod.rs
katat: oh right!

src/mast/mod.rs
katat: looks like it will need to have a check to see if all the generics are instantiated before this.

src/parser/types.rs
katat: right, we would need to also rely on the GenericArray to carry that constant variable.
I guess this is another reason why we need to refactor the `Array` to carry symbolic for both concrete Array and generic Array.

katat: (as you pointed out in the main PR)

src/stdlib/mod.rs
katat: these builtins don't need generics.
but the others would need them, such as bit functions.

src/type_checker/checker.rs
katat: this todo is outdated. 
the extraction is already done in the `FnSig`

src/type_checker/mod.rs
katat: The generics store parameters by their unique names.

If the hash map is based on the `Ident`, it can be difficult to point to the right one?
such as 
```rust
fn arr(aa: [[Field; N]; N])
```

src/mast/mod.rs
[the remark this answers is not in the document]
mimoo: usually you want to have

```rust
struct GenericParameters { /* ... */ }
impl GenericParameters { /* ... */ } // <-- the impl right after 
```

src/mast/mod.rs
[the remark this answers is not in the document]
mimoo: yeah I think that's better

src/mast/mod.rs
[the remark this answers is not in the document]
mimoo: but is it?

src/mast/mod.rs
mimoo: I think the parser can do that check no?

mimoo: yeah that seems unrelated to the MAST

src/type_checker/mod.rs
mimoo: oh I see, good point!

src/mast/mod.rs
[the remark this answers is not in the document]
katat: I was trying to avoid importing `MonomorphizedFnEnv` in the `parser/types.rs`.
It is a bit weird the `parser/types.rs` depends on something relevant to monomorphization.

src/mast/mod.rs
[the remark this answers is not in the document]
katat: Good point. I should list them in the RFC.

src/mast/mod.rs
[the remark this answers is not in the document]
mimoo: yeah agree!

PR #154, opened 2024-07-29
Add `to_bits` and `from_bits`  builtin functions

katat:
todo:
- [ ] support importing module, such as `use std::bits` instead of `use std::to_bits`
- [ ] reconsider the use of `Backend::add_constant`


katat:
To discuss:

The generic values are currently passed from the `FnSig::generics` to the corresponding built-in function. However, directly passing in the concrete types for the resolved generic arguments seems cleaner. 

The concrete types not only contains all the values that the `FnSig::generics` carries, but also the structure. So the built-in function doesn't need to load the generic argument and evaluate the generic type with the passed generic values again.

katat:
Actually, based on the latest RFC, it can generate a "instantiated fully qualified" string as a key to load the monomorphized `FunctionDef`, which contains the resolved concrete types.

So this may be not a big deal, since the `generics` argument for built-in functions can be used to generate the postfix for the key string. The current `generics` argument might be also handy for some cases, where we don't want to manually extract the generic value from a type.

src/stdlib/mod.rs
```diff
@@ -163,3 +169,127 @@ fn assert<B: Backend>(
...
+            .mul(&bit, &bit_minus_one, Span::new(2, 3, 1));
+        compiler
+            .backend
+            .assert_eq_const(&zero_mul, zero, Span::new(2, 4, 1));
+
+        // lc += bits[i] * e2;
+        let weighted_bit =
+            compiler
+                .backend
+                .mul_const(&bit, &B::Field::from(e2), Span::new(2, 5, 1));
```
mimoo, line 234: why these fake spans btw :D?

src/stdlib/mod.rs
```diff
@@ -163,3 +169,127 @@ fn assert<B: Backend>(
...
+
+    // second var is the value to convert
+    let val = match &var[0] {
+        ConstOrCell::Cell(cvar) => cvar,
+        ConstOrCell::Const(_) => todo!(),
+    };
+
+    // convert value to bits
+    let mut bits = Vec::new();
+    let mut e2 = 1u32;
```
mimoo, line 203: `Field::one()` otherwise it will overflow

src/stdlib/mod.rs
```diff
@@ -163,3 +169,127 @@ fn assert<B: Backend>(
...
+    span: Span,
+) -> Result<Option<Var<B::Field, B::Var>>> {
+    // should be two input vars
+    assert_eq!(vars.len(), 2);
+
+    // but the better practice would be to retrieve the value from the generics
+    let num_var = generics.get("LEN");
+
+    // num should be greater than 0
+    assert!(num_var > 0);
```
mimoo, line 186: let's also assert that `bitlen <= MODULUS.bitlen() - 1`

src/stdlib/mod.rs
mimoo: (as `u32` is 32-bit, but we could have values up til ~250 bits in some of these curves)

src/stdlib/mod.rs
```diff
@@ -163,3 +169,127 @@ fn assert<B: Backend>(
...
+) -> Result<Option<Var<B::Field, B::Var>>> {
+    // only one input var
+    assert_eq!(vars.len(), 1);
+
+    let var_info = &vars[0];
+    let num_bits = generics.get("LEN");
+
+    let bits_vars = var_info.var.cvars.iter().map(|c| match c {
+        ConstOrCell::Cell(c) => c,
+        ConstOrCell::Const(_) => todo!(),
```
mimoo, line 271: we should address this TODO in this PR

src/stdlib/mod.rs
```diff
@@ -163,3 +169,127 @@ fn assert<B: Backend>(
...
+    });
+
+    // this might not be necessary since it should be checked in the type checker
+    assert_eq!(num_bits as usize, bits_vars.len());
```
mimoo, line 275: doesn't hurt I guess :D BTW `num_bits` is already a usize

src/stdlib/mod.rs
```diff
@@ -163,3 +169,127 @@ fn assert<B: Backend>(
...
+
+    let mut e2 = 1u32;
```
mimoo, line 277: same issue here

src/var.rs
```diff
@@ -66,6 +78,7 @@ impl<B: Backend> std::fmt::Debug for Value<B> {
             Value::External(..) => write!(f, "External"),
             Value::PublicOutput(..) => write!(f, "PublicOutput"),
             Value::Scale(..) => write!(f, "Scaling"),
+            Value::RSAnd(_, _) => todo!(),
```
mimoo, line 81: ?

src/var.rs
```diff
@@ -46,6 +46,18 @@ where
...
+    /// Extract the right most bit after an arithmetic right shift on the given variable.
+    /// e.g. `value >> 1 & 1`
+    // todo: consider using Value itself as argument, which wraps B::Var or B::Field, as Value::Var or Value::Field
+    // - so the arguments for these operations can be either B::Var or B::Field
+    // - there can be some cases where the value to shift doesn't need to be from the witness vector
+    // - for example, when adding a symbolic value to represent the `value >> 1 & 1`,
+    // - we simply want to calculate the value without needing to add cell var for the intermediate value, such as `value >> 1`
+    // - the problem of adding the cell var for the intermediate value is that
+    // - the var likely won't end up in the circuit, that would trigger error.
+    // but does this make sense to all different backends? is it possible that some backend doesn't allow certain out of circuit calculations like this?
```
mimoo, line 58: not sure I understand this comment... maybe we can talk about it

src/tests/examples.rs
```diff
@@ -562,3 +562,21 @@ fn test_generic_for_loop(#[case] backend: BackendKind) -> miette::Result<()> {
 
     Ok(())
 }
+
+#[rstest]
+#[case::kimchi_vesta(BackendKind::KimchiVesta(KimchiVesta::new(false)))]
+#[case::r1cs(BackendKind::R1csBls12_381(R1CS::new()))]
+fn test_generic_builtin_bits(#[case] backend: BackendKind) -> miette::Result<()> {
+    let public_inputs = r#"{"xx":"2"}"#;
+    let private_inputs = r#"{}"#;
```
mimoo, line 571: I'm thinking more and more that we should have the inputs to these tests closer to each files. Maybe we could start our files with the inputs as comments? For example `generic_builtin_bits.no`:

```rust
// public_inputs = {"xx":"2"}
// private_inputs = {}
fn main(_) { ... }
```

mimoo:
It would be nice to move this to a `std::bits` module as well

src/stdlib/mod.rs
```diff
@@ -29,9 +29,13 @@ pub const QUALIFIED_BUILTINS: &str = "std/builtins";
 
 const ASSERT_FN: &str = "assert(condition: Bool)";
 const ASSERT_EQ_FN: &str = "assert_eq(lhs: Field, rhs: Field)";
+const TO_BITS_FN: &str = "to_bits(const LEN: Field, val: Field) -> [Bool; LEN]";
+const FROM_BITS_FN: &str = "from_bits(bits: [Bool; LEN]) -> Field";
```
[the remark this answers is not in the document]
mimoo: https://github.com/zksecurity/noname/issues/155

src/stdlib/mod.rs
katat: It was for debugging. I should remove them.

src/stdlib/mod.rs
katat: Added.
But why it is not `bitlen <= MODULUS.bitlen()`?

katat:
that is strange to see the tests failed on CI, but it works on my machine. 
can you reproduce this on your machine? @mimoo

src/stdlib/mod.rs
mimoo: Imagine a modulus that’s not 1111, but like 1001 for example

mimoo:
I can't compile on my machine, I get a number of compilation issues:

```
error[E0599]: no method named `get` found for reference `&GenericParameters` in the current scope
   --> src/stdlib/mod.rs:264:27
    |
264 |     let bitlen = generics.get("LEN") as usize;
    |                           ^^^ method not found in `&GenericParameters`
    |
    = help: items from traits can only be used if the trait is implemented and in scope
    = note: the following traits define an item `get`, perhaps you need to implement one of them:
            candidate #1: `SliceIndex`
            candidate #2: `toml_edit::table::TableLike`
help: one of the expressions' fields has a method of the same name
    |
264 |     let bitlen = generics.0.get("LEN") as usize;
    |                           ++
```

mimoo:
can't reproduce either. Is it possible that there's a source of non-determinism? Maybe a hashmap somewhere?

examples/generic_builtin_bits.no
```diff
@@ -1,13 +1,23 @@
...
     assert(!bits[2]);
 
     let val = from_bits(bits);
     assert_eq(val, xx);
+
+    // constant
+    let cst_bits = to_bits(3, 2);
+    assert(!cst_bits[0]);
+    assert(cst_bits[1]);
+    assert(!cst_bits[2]);
```
mimoo, line 19: it'd be good to also `assert(cst_bits.len() == 3)`

thinking about that, I never liked that in circom you can write asserts that don't encode anything. So we may want to forbid `assert()` on a constant and have a `assert_dbg(cst_bits.len() == 3)` instead

mimoo: https://github.com/zksecurity/noname/issues/156

src/stdlib/mod.rs
```diff
@@ -180,75 +182,70 @@ fn to_bits<B: Backend>(
     assert_eq!(vars.len(), 2);
 
     // but the better practice would be to retrieve the value from the generics
-    let num_var = generics.get("LEN");
+    let bitlen = generics.get("LEN") as usize;
```
mimoo, line 185: also can we move all of these to a `bits.rs` file?

mimoo (approved):
\o/

src/stdlib/mod.rs
katat: so if the modulus is 9 (1001), the `MODULUS.bitlen() - 1` will only allow it represent max value as 111, which is 7.

then it doesn't work for the value 8?

mimoo: how would you prove that like 10, 11, 12, 13, 14, and 15 are correctly constrained?

Issue #157, opened 2024-07-31
handle modules better


mimoo:
This is a main issue for a number of issues we need to fix.

## Enforcing allowed imports

First, I can import anything I want here:

```rust
use something::that_doesnt_exist;
```

and it will work. This is because the name resolution pass doesn't seem to be aware of dependencies that are allowed. The `Noname.toml` file related to the code being compiled should dictate which dependencies are accessible. Some standard modules should be accessible as well (not `std::builtins` which is included by default, but `std::crypto` or `std::bits`)

## Scope of imports

Second, it seems to me that we aren't really keeping track of what dependencies a function has access to as we traverse the AST. This is because we build one main TAST with all the fully-qualified functions (or exported structs or constants) in  its `functions: HashMap<FullyQualified, FnInfo<B>>` field when we compile all dependencies in `cmd_build_and_check.rs:produce_all_asts()`:

```rust
for dep in dep_graph.from_leaves_to_roots() {
        let path = path_to_package(&dep);

        let lib_file = path.join("src").join("lib.no");
        let code = std::fs::read_to_string(&lib_file)
            .into_diagnostic()
            .wrap_err_with(|| format!("could not read file `{path}`"))?;

        let lib_file_str = lib_file.to_string();
        node_id = typecheck_next_file(
            &mut tast,
            Some(dep),
            &mut sources,
            lib_file_str.clone(),
            code,
            node_id,
        )?;
    }
```

This means that if we have the following chain of imports `a -> b -> c` then `a` will be able to call `c`'s functions. Worst, even parts of the dependency graph might have access to their neighboring parts if they are parsed after. I think we should keep a dependency graph on the side AND go through it as we go through different modules, so that we can keep track of what other modules are accessible to us.

I'm wondering if this could be done in the name resolution phase as explained in the previous section (so basically I might be repeating the same stuff here).

## Move to declaring stdlib modules in noname files!

https://github.com/zksecurity/noname/issues/155

PR #158, opened 2024-07-31
proper modules

mimoo:
What I did in this PR:

* move bits and builtins stuff in their own rust modules
* introduce `Module` trait so that each module can just implement that

This is still a bit ugly (why enum lol) but it should be fine as it should be temporary. (because ideally modules should be declared as their own files, see https://github.com/zksecurity/noname/issues/157)

This will allow us to focus on implementing things instead of improving the modules. For example, if we want to mix native and builtins we can have builtins live in `std::bits_builtins` and the native code in `std::bits`.

### DIFF
--- a/src/stdlib/mod.rs
+++ b/src/stdlib/mod.rs
@@ -1,310 +1,80 @@
-use std::collections::HashSet;
-
-use ark_ff::{One, Zero};
-use kimchi::o1_utils::FieldHelpers;
-use once_cell::sync::Lazy;
-
 use crate::{
     backends::Backend,
     circuit_writer::{CircuitWriter, VarInfo},
     constants::Span,
-    constraints::boolean,
-    error::{Error, ErrorKind, Result},
+    error::Result,
     imports::FnKind,
     lexer::Token,
     parser::{
-        types::{FnSig, GenericParameters, TyKind},
+        types::{FnSig, GenericParameters},
         ParserCtx,
     },
     type_checker::FnInfo,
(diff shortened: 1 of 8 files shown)


src/stdlib/mod.rs
```diff
@@ -1,310 +1,80 @@
...
-    let rhs_info = &vars[1];
-
-    // they are both of type field
-    if !matches!(lhs_info.typ, Some(TyKind::Field | TyKind::BigInt)) {
-        panic!(
-            "the lhs of assert_eq must be of type Field or BigInt. It was of type {:?}",
-            lhs_info.typ
-        );
+impl AllStdModules {
+    pub fn all_std_modules() -> Vec<AllStdModules> {
```
katat, line 27: is `new` better as a constructor name?

katat: Would using dynamic dispatch be better? like:

```rust
pub fn all_std_modules() -> Vec<Box<dyn Module>> {
    vec![
        Box::new(BuiltinsModule),
        Box::new(CryptoModule),
        Box::new(BitsModule),
    ]
}
```

But it can be tricky though. I remember it was difficult do something like this when it involves type parameters.

katat (approved):
Looks great!

katat (approved):
Looks great!

src/stdlib/mod.rs
mimoo: I tried! but failed :D

PR #161, opened 2024-08-22
Remove $ characters from commands

microbecode:
If user is an idiot, like me, clicking the "copy" button next to the commands copies also the $ character, which obviously makes the command not work.

Removed the $ characters so copypasting is easier.

Also changed the tree root to be called "root" instead of "tree".

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ fn main(pub public_input: Field, private_input: Field) -> Bool {
 You can run the above example with the following command:
 
 ```console
-$ noname test --path examples/public_output_bool.no --private-inputs '{"private_input": "1"}' --public-inputs '{"public_input": "1"}' --debug
+noname test --path examples/public_output_bool.no --private-inputs '{"private_input": "1"}' --public-inputs '{"public_input": "1"}' --debug
 ```
 
 On particularity of noname is the `--debug` option that shows you how the code relates to the compiled constraints:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -57,15 +57,15 @@ We have a lot of resources to learn and understand how noname works:
...
-$ noname new --path my_project
+noname new --path my_project
 ```
 
 This will create a `Noname.toml` manifest file, which contains the name of your project (which must follow a Github `user/repo` format) as well as dependencies you're using (following the same format, as they are retrieved from Github).
 
 This will also create a `src` directory, which contains a `main.no` file, which is the entry point of your program. If you want to create a library, pass the `--lib` flag to the `new` or `init` command of `noname`, and it will create a `lib.no` file instead.
 
 ```
-$ tree
```
katat, line 68: I think the `tree` indicates a command to list the folder content :D

katat (commented):
Thanks for the PR @microbecode 
Yeah, it makes sense to me to remove the $ 

cc @mimoo

README.md
microbecode: Adjusted now.

mimoo:
usually the `$` sign is pretty standard to show that it's a terminal command, also I'm not sure what `root` is here. I don't think this change is impactful enough, but I appreciate the contribution. Instead I would redirect you to check the list of issues that we have here that are sorted from easy to hard: https://github.com/zksecurity/noname/issues !

PR #163, opened 2024-08-22
Add constant field to TyKind::BigInt

katat:
This a quick experimental change to allow to type check the const attribute in function.

This might be a good chance to complete this [todo note](https://github.com/zksecurity/noname/blob/3f08ecef82ef47922afaaeab37d63066d5c38dfd/src/parser/types.rs#L233).

By adding constant field to `TyKind::Field` instead, we might be able to remove the `TyKind::BigInt`, which is easy to confuse with `TyKind::Field`.


katat:
go with another approach

Issue #151, opened 2024-07-23
Document Bigint to figure out if we can remove it


mimoo:
https://github.com/zksecurity/noname/blob/10a24bdc2078371e5f52f8cf5c266fcbd9b4d41e/src/parser/types.rs#L184

![Screenshot 2024-07-22 at 8 38 48 PM](https://github.com/user-attachments/assets/f17790da-da09-46a2-9255-9366b05ec29c)

I can't remember why we had this type, it would be good to spend some time tracking it through the codebase and document exactly what is its purpose. This way we can decide if we want to remove it, or refactor things.

ronantakizawa:
BigInt is used in the code to handle very large signed integers that can't be represented by standard fixed-size integer types.

Perhaps we could use Fixed-Size Integer Types or Fixed-size finite field elements

mimoo:
I believe the PR mentioned above deletes it :o

ronantakizawa:
The above PR looks good 👍

Issue #22, opened 2024-03-25
can we get rid of the `const` keyword?


mimoo:
check the [iterate.no](https://github.com/zksecurity/noname/blob/main/examples/iterate.no) example:

```rust
fn House.room(self, const idx: Field) -> Room {
    return self.rooms[idx];
}
```

do we really need the `const` here? Can't we just have the compiler complain if you pass a variable instead of a const for an actual circuit? It feels to me like this keyword just adds more mental load to the developer

mimoo:
in my notes I had "replace `const Field` with `Number` or `Int` or smthg"

mimoo:
I think the const keyword is acting like a generic right now, which is nicer than having an extra syntax like `fn House.room[idx](self)` IMO

PR #164, opened 2024-08-23
refactor TyKind::Field to contain constant flag

katat:
This aims to resolve for the issues https://github.com/zksecurity/noname/issues/22 https://github.com/zksecurity/noname/issues/151

- As the added negative test demonstrated, the `const` attribute is important to have. It help prevent misusing input fields for function arguments that are supposed to be constant.
- A `constant` field is added to `TyKind::Field`.
- `TyKind::BigInt` is deprecated, since the `TyKind::Field` itself can differentiate if it is a constant.
- Deprecated the `TypeInfo.constant`, since the `TyKind::Field` itself can propagate the constant info.


src/type_checker/checker.rs
```diff
@@ -378,7 +369,7 @@ impl<B: Backend> TypeChecker<B> {
                 // check that expression is a bigint
                 let idx_typ = self.compute_type(idx, typed_fn_env)?;
                 match idx_typ.map(|t| t.typ) {
-                    Some(TyKind::BigInt) => (),
+                    Some(TyKind::Field { constant: true }) => (),
```
mimoo, line 372: btw probably you should have targeted the new const generic branch :D as you will have to add more changes for default array init also

src/type_checker/checker.rs
```diff
@@ -596,8 +587,10 @@ impl<B: Backend> TypeChecker<B> {
                 typed_fn_env.nest();
 
                 // create var (for now it's always a bigint)
```
mimoo, line 589: obsolete comment now

src/type_checker/checker.rs
```diff
@@ -682,6 +675,14 @@ impl<B: Backend> TypeChecker<B> {
 
         // compare argument types with the function signature
         for (sig_arg, (typ, span)) in expected.iter().zip(observed) {
+            // when const attribute presented, the argument must be a constant
+            if sig_arg.is_constant() && !matches!(typ, TyKind::Field { constant: true }) {
+                return Err(self.error(
+                    ErrorKind::ArgumentTypeMismatch(sig_arg.typ.kind.clone(), typ),
+                    span,
+                ));
+            }
```
mimoo, line 684: Screenshot 2024-08-23 at 3 18 57 PM

^ I managed to write this and it compiles :D perhaps we should add a check that we can only have `const Field` arguments in a function, and not `const CustomType`. If we want to generalize it one day we can do it, but at least we don't have to do the work now, wdyt?

mimoo: it's also weird that now we have `arg.is_constant()` which checks the `FnArg.attribute`. Do we even need this to track constants now? It seems like the `match_expected` call right below this comment would be enough (if you address my earlier comment)

mimoo: my recommendation would be to keep both checks at this point, but add a comment that this might be unnecessary :D

src/type_checker/mod.rs
```diff
@@ -208,7 +207,7 @@ impl<B: Backend> TypeChecker<B> {
                             ConstInfo {
                                 value: vec![cst.value],
                                 typ: Ty {
-                                    kind: TyKind::Field,
+                                    kind: TyKind::Field { constant: true },
```
mimoo, line 210: Some thoughts: in theory we can have constants of any type. But now we only support `Field` constants. 

Also, a non-`Field` value can easily be checked to be a constant or not by checking if all of its `Field` fields are constants (because down the line a type is constituted of many `Field`s). So it should be straight forward to support non-`Field` consts at some point.

I don't think it's urgent though, so I will not create an issue for that

mimoo (approved):
beautiful! I like how this simplifies a lot of the code.

src/parser/types.rs
```diff
@@ -200,7 +195,7 @@ pub enum TyKind {
 impl TyKind {
     pub fn match_expected(&self, expected: &TyKind) -> bool {
         match (self, expected) {
-            (TyKind::BigInt, TyKind::Field) => true,
+            (TyKind::Field { .. }, TyKind::Field { .. }) => true,
```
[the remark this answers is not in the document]
katat: If `match_expected` is a relaxed version of `same_as`, and the lhs and rhs doesn't matter, then it seems the following aligns with the intention:
```rust
(TyKind::Field { .. }, TyKind::Field { .. }) => true,
```

In practice, this relaxed check for field type seems necessary. Here are the cases where the `match_expected` and `same_as` are applied to check for the field type.

Assignment case:
```rust
fn main(pub xx: Field) {
  let mut yy = 2; // constant
  yy = xx; // should allow assigning a var value
}
```

Argument case:
```rust
allow either constant or non constant argument
fn add(xx: Field, yy: Field) -> Field {
  return xx + yy; // should allow binary ops over the mix of constants or non constants
}

// yy only allows constant value. 
fn add_const(xx: Field, const yy: Field) -> Field {
 return xx + yy; 
}

fn main(pub xx: Field) {
  // first arg is var, while 2 is a constant
  add(xx, 2);
}
```

Custom type declaration case:
```rust
Thing {
  xx: Field // allow either constant or not
}
fn main(pub xx: Field) {
  // based on var
  let thing1 = Thing {xx: xx};
  // based on constant
  let thing2 = Thing {xx: 2};
}
```

It is indeed getting a bit weird to see the `same_as` function also allows relaxed checking for the Field type though.

Maybe we should rename the `match_expected` to `match_generic`.

Or maybe we should add a bunch of functions to specify individually for each of these cases, instead of using an unified rule for all the cases listed above. For example:
- check_assignment(lhs_typ, rhs_typ)
- check_argument(observed_typ, expected)
- check_return(observed, expected)

This may make it clearer the context in which what exceptions can be made, such as for the Field case.

src/parser/types.rs
```diff
@@ -218,7 +213,6 @@ impl TyKind {
 
     pub fn same_as(&self, other: &TyKind) -> bool {
```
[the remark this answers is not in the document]
katat: Maybe what we really need is to make specific cases for comparing types :D
See my comment above.

src/parser/types.rs
```diff
@@ -381,7 +377,16 @@ impl FnSig {
...
+        let mut arguments = FunctionDef::parse_args(ctx, tokens, &kind)?;
+
+        // if it is with const attribute, then converts it to a constant field.
+        // this is because the parser doesn't know if a token has a corresponding attribute
+        // until it has parsed the whole token.
+        for arg in &mut arguments {
+            if arg.is_constant() {
+                arg.typ.kind = TyKind::Field { constant: true };
+            }
+        }
```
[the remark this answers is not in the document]
katat: good point!
the code is shifted to here.

src/type_checker/checker.rs
katat: agreed, I have just change this PR's base branch to the const generic branch.

src/type_checker/checker.rs
katat: I think the `is_constant` is still necessary. This is because for the function signature
- with const attribute, it only accept constant field.
- without const attribute, it accept the field type in either mode.

As I mentioned above, it may be better to create a specific checking rule for function signature, so as to make the rules clearer for different cases and avoid being confused by the `match_expected` function.

katat: updated to forbid const over types other than field type

Issue #166, opened 2024-08-23
document how match_expected is different from same_as


mimoo:
we have two functions:

* match_expected
* same_as

and they seem to be doing the same thing, almost. There's no documentation and I don't remember why I wrote them like this. It would be helpful if someone can investigate why we have two functions here, and either merge them into one or document why we need two different functions

ronantakizawa:
Hi @mimoo , this is my first issue on Noname so I may have some misunderstandings about the code. 

From my understanding these are the differences: 

`match_expected`: Allows for flexible matching between types, meaning it considers types equivalent in certain contexts, such as treating BigInt as equivalent to Field.

`same_as`: Requires an exact match between types, ensuring that two types are completely identical.

mimoo:
Hey! So what we're wondering is how is that difference used? Why is one called over another? (Like who calls the first one, who calls the second one)

Especially after this PR: https://github.com/zksecurity/noname/pull/164

I think a PR merging both of the functionality with some documentation and a new argument flag that allows switching between the two behaviors would be nice!

PR #167, opened 2024-08-26
Merge type-checking into 1 function

ronantakizawa:
My first PR let me know if there are contribution practices that I missed!

`match_expected` and `same_as` are merged into the function `compare_with` which has a flag that lets you control if you want an exact match or not. Included changes in docs (type-checker.md).

This PR includes changes from PR #164. 

Resolves issue #166


book/src/type-checker.md
```diff
@@ -66,7 +66,33 @@ So that shadowing is disallowed, even in different scopes, there is only one var
...
+```
+if !rhs_typ.typ.compare_with(&lhs_node.typ, true) {  // Change `true` to `false` if leniency is required
+                    return Err(self.error(
+                        ErrorKind::MismatchType(lhs_node.typ.clone(), rhs_typ.typ.clone()),
+                        expr.span,
+                    ));
+                }
+
+```
+In this example, the function checks whether the right-hand side type (`rhs_typ.typ`) matches the left-hand side type (`lhs_node.typ`) with leniency. If they do not match, an error is returned.
```
mimoo, line 91: oh I meant some documentation for the function in the code :D I don't think we need to explain the behavior of the function in the book

src/parser/types.rs
```diff
@@ -193,38 +193,34 @@ pub enum TyKind {
...
-                    name: expected_name,
-                },
-            ) => module == expected_module && name == expected_name,
-            (x, y) if x == y => true,
-            _ => false,
-        }
-    }
-
-    pub fn same_as(&self, other: &TyKind) -> bool {
+    pub fn compare_with(&self, other: &TyKind, exact_match: bool) -> bool {
```
mimoo, line 196: maybe `matches(_)` would be a better/shorter name?

src/parser/types.rs
```diff
@@ -193,38 +193,34 @@ pub enum TyKind {
...
-            ) => module == expected_module && name == expected_name,
-            (x, y) if x == y => true,
-            _ => false,
+            ) => module == other_module && name == other_name,
+    
+            // Exact match comparison
+            (x, y) if exact_match => x == y,
+    
+            // Fallback to basic equality check
+            (x, y) => x == y,
```
mimoo, line 223: `(x, y) => x == y` should be enough for the last two branches

src/parser/types.rs
```diff
@@ -193,38 +193,34 @@ pub enum TyKind {
...
-
-    pub fn same_as(&self, other: &TyKind) -> bool {
+    pub fn compare_with(&self, other: &TyKind, exact_match: bool) -> bool {
         match (self, other) {
+            // Handle the Field types with optional exact match logic
+            (TyKind::Field { constant: true }, TyKind::Field { constant: false })
+            | (TyKind::Field { constant: false }, TyKind::Field { constant: true }) => {
+                !exact_match
+            }
+            (TyKind::Field { .. }, TyKind::Field { .. }) => true,
```
mimoo, line 203: so I think this is the tricky part, I actually think we should write this:

```rust
(TyKind::Field { constant: observed }, TyKind::Field { constant: expected }) =>
  (!expected || observed) 
```

and this assume that the expected/signature is the `other` and the one we're checking against it is `self` which is not necessarily implied by the name of the function. So perhaps we should keep `match_expected` as a name

src/type_checker/checker.rs
```diff
@@ -394,7 +394,7 @@ impl<B: Backend> TypeChecker<B> {
                         .expect("expected a value");
 
                     if let Some(tykind) = &tykind {
-                        if !tykind.same_as(&item_typ.typ) {
+                        if !tykind.compare_with(&item_typ.typ,true) {
```
mimoo, line 397: so `same_as` is only used in a single place, but it doesn't make sense to me. Why wouldn't we be able to declare an array with a mix of constants and non-constants? We might want to get rid of the `same_as` function altogether

mimoo (commented):
Thanks for the PR! Can you target https://github.com/zksecurity/noname/pull/164 instead of main?

also @katat what do you think of my comment? I think we should just get rid of `same_as` and use a strong check only when we're expecting a constant. But it'd be good if someone double checked my rationale ^^

src/type_checker/checker.rs
mimoo: actually, I'm backtracking here, I think for array declaration we'll want to have a flag that says we don't even care about the constant in a Field, maybe the flag could be called `ignore_constants`

katat:
I align with your observations on the constant / non-constant @mimoo . Actually I try to detail my ideas on how to resolve this `match_expected` and `same_as` confusion: https://github.com/zksecurity/noname/pull/164#discussion_r1732105951 :D

src/parser/types.rs
```diff
@@ -292,44 +292,50 @@ pub enum TyKind {
...
+    /// # Returns
+    /// - `true` if the `self` and `expected` types match according to the rules defined and the `ignore_constants` flag.
+    /// - `false` otherwise.
+    ///
+    /// # Matching Rules
+    /// - If both types are `Field`, the function compares their `constant` fields unless `ignore_constants` is `true`.
+    /// - For `Array` types, it checks if both the size and element types match recursively.
+    /// - For `Custom` types, it compares the `module` and `name` fields for equality.
+    /// - For other types, it uses basic equality check
+    pub fn match_expected(&self, expected: &TyKind, ignore_constants: bool) -> bool {
```
katat, line 310: `ignore_constants` doesn't seem to apply to anything. :D

katat (commented):
I think it is good to conslidate these two functions into one.

As my rationale in the other PR's comment https://github.com/zksecurity/noname/pull/164#discussion_r1732105951, later we can decide which set of rules to be individualized if the rules started getting too complex in different contexts for one function to generalize.

src/type_checker/checker.rs
```diff
@@ -258,7 +258,7 @@ impl<B: Backend> TypeChecker<B> {
                 // and is of the same type as the rhs
                 let rhs_typ = self.compute_type(rhs, typed_fn_env)?.unwrap();
 
-                if !rhs_typ.typ.match_expected(&lhs_node.typ) {
+                if !rhs_typ.typ.match_expected(&lhs_node.typ, true) {
```
katat, line 261: This type check works because when it comes to assignment expression the assumption is that the variable is under mutable mode.
So it should allow assigning either constant or non constant field to the variable.

src/type_checker/checker.rs
```diff
@@ -409,7 +409,7 @@ impl<B: Backend> TypeChecker<B> {
                         .expect("expected a value");
 
                     if let Some(tykind) = &tykind {
-                        if !tykind.same_as(&item_typ.typ) {
+                        if !tykind.match_expected(&item_typ.typ, false) {
```
katat, line 412: It uses same_as is because it expects the types are not generic and can be matched. The same_as function only consider the case for resolved types.

Also in this case, it doesn't have to be constant to constant match. I think it should allow an array defined as 
```rust
let arr = [var1, 1];
```

So it seems we still need `same_as` here. Or maybe the `match_expected` can have another flag `generic_not_allowed`

src/type_checker/checker.rs
```diff
@@ -507,7 +507,7 @@ impl<B: Backend> TypeChecker<B> {
                         .compute_type(&observed.1, typed_fn_env)?
                         .expect("expected a value (TODO: better error)");
 
-                    if !observed_typ.typ.match_expected(&defined.1) {
+                    if !observed_typ.typ.match_expected(&defined.1, true) {
```
katat, line 510: Similar to the comment to the assignment case above, this is because it allows assigning either constant or not.

src/type_checker/checker.rs
```diff
@@ -587,7 +587,7 @@ impl<B: Backend> TypeChecker<B> {
                 return Err(self.error(ErrorKind::NoReturnExpected, stmts.last().unwrap().span))
             }
             (Some(expected), Some(observed)) => {
-                if !observed.match_expected(&expected.kind) {
+                if !observed.match_expected(&expected.kind, true) {
```
katat, line 590: This allows constant because the value being returned could be either constant or not. For example:
```rust
return arr[idx]; // the element behind the idx could be either a field or a constant field.
```

src/type_checker/checker.rs
katat: cc @mimoo

src/type_checker/checker.rs
```diff
@@ -727,7 +727,7 @@ impl<B: Backend> TypeChecker<B> {
                 ));
             }
 
-            if !typ.match_expected(&sig_arg.typ.kind) {
+            if !typ.match_expected(&sig_arg.typ.kind, true) {
```
katat, line 730: the arguments can be either constant or not, unless the signature specifies the const attribute.

src/mast/mod.rs
```diff
@@ -496,7 +496,7 @@ fn monomorphize_expr<B: Backend>(
             let lhs_typ = lhs_mono.typ.unwrap();
             let rhs_typ = rhs_mono.typ.unwrap();
 
-            if !lhs_typ.same_as(&rhs_typ) {
+            if !lhs_typ.match_expected(&rhs_typ, true) {
```
katat, line 499: Same as the other comment, This may require another flag to ensure the types are resolved types and not generic involved.

src/mast/mod.rs
```diff
@@ -684,7 +684,7 @@ fn monomorphize_expr<B: Backend>(
                 items_mono.push(item_mono.clone());
 
                 if let Some(tykind) = &tykind {
-                    if !tykind.same_as(&item_mono.clone().typ.unwrap()) {
+                    if !tykind.match_expected(&item_mono.clone().typ.unwrap(), true) {
```
katat, line 687: same

src/mast/mod.rs
```diff
@@ -758,7 +758,7 @@ fn monomorphize_expr<B: Backend>(
                 let observed_mono = &monomorphize_expr(ctx, &observed.1, mono_fn_env)?;
                 let typ_mono = observed_mono.typ.as_ref().expect("expected a value");
 
-                if !typ_mono.same_as(&defined.1) {
+                if !typ_mono.match_expected(&defined.1, true) {
```
katat, line 761: same

src/mast/mod.rs
```diff
@@ -840,7 +840,7 @@ pub fn monomorphize_block<B: Backend>(
 
     // check the return
     if let (Some(expected), Some(observed)) = (expected_return, return_typ.clone()) {
-        if !observed.same_as(&expected.kind) {
+        if !observed.match_expected(&expected.kind, true) {
```
katat, line 843: same

katat (commented):
I think we need to add another flag `no_generic_allowed` to match_expected, so that it throws when unexpected generic types presented.

I added a couple comment to remark why the constant can be ignored at certain places. Maybe we can add these as comments to the uses of `match_expected`, so people know why sometimes the constant in the type check can be ignored.

Also the only place the constant check is required seems to be when there is a const attribute for a function argument. I made a comment to the `ArrayDeclaration` case on why it might not need constant check. If this is the case, then we might not really need flag `ignore_constant`.

ronantakizawa:
I experimented with using `ignore_constants` with `no_generics_allowed` and just `no_generics_allowed` and `ignore_constants` wasn't doing anything so I removed it.

katat:
Nice work! 
This PR, including both the code and the conversations, really help consolidate one of the tricky parts in the type checking process.

Issue #83, opened 2024-06-01
create a test for snarkjs


mimoo:
should we create a CI test for this: https://gist.github.com/katat/7482d07ae783a5b20a4e5f04dec4cb16 ?

katat:
We should! I think we can dockerize the script and update the github action to run it.

katat:
Here is the tutorial about the workflow: https://www.zksecurity.xyz/blog/posts/noname-r1cs/

PR #168, opened 2024-08-27
Add Snarkjs test to CI

ronantakizawa:
Add Snarkjs test to CI. Resolves issue #83. 

Tests based on this shell script: https://gist.github.com/katat/7482d07ae783a5b20a4e5f04dec4cb16

### DIFF
--- a/.github/workflows/Dockerfile
+++ b/.github/workflows/Dockerfile
@@ -1,0 +1,25 @@
+# Install rust 1.65 for Circom
+FROM rust:1.65.0 as circom-instaler
+
+RUN git clone https://github.com/iden3/circom.git && \
+    cd circom && \
+    cargo build --release && \
+    cargo install --path circom
+
+# Install Node.js 16 and npm
+RUN curl -fsSL https://deb.nodesource.com/setup_16.x | bash - \
+    && apt-get install -y nodejs \
+    && npm install -g snarkjs \
+    && apt-get clean && rm -rf /var/lib/apt/lists/*
+
+# Set the working directory
+WORKDIR /app
+
+# Copy the script into the container
+COPY snarkjs.sh /app/snarkjs.sh
+
+# Ensure the script has execution permissions
+RUN chmod +x /app/snarkjs.sh
+
+# Define the entry point to run the script
+CMD ["/app/snarkjs.sh", "/app", "bn128", "--new"]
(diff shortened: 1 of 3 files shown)


.github/workflows/snarkjs.sh
```diff
@@ -0,0 +1,107 @@
...
+
+# Check if directory path argument is provided
+if [ "$#" -lt 2 ]; then
+    echo "Usage: $0 <directory_path> <curve>"
+    exit 1
+fi
+
+DIR_PATH=$1
+CURVE=$2
+NEW_PTAU=false
```
katat, line 11: we will need to download the ptau from somewhere, if it is too time consuming to rebuild on the CI.

.github/workflows/snarkjs.sh
```diff
@@ -0,0 +1,107 @@
...
+
+# Create a basic input.json for testing
+if [ ! -f "$DIR_PATH/input.json" ]; then
+    echo "Generating a basic input.json file for testing..."
+    cat > "$DIR_PATH/input.json" <<EOL
+{
+    "a": 3,
+    "b": 11
+}
+EOL
```
katat, line 44: I am not sure what is this for. 
I think we need to test against the circuits generated by noname instead of circom.

katat (commented):
Thanks for the PR. This gonna to be very useful. We need more this kind of integration tests!

The only comment I have is to make the docker container to generate circuits from the noname CLI instead of circom CLI. Here is a tutorial about the workflow: https://www.zksecurity.xyz/blog/posts/noname-r1cs/

ronantakizawa:
Thanks for the feedback @katat 
I wasn't too familiar with the noname CLI but I use it in the CI now. 

I couldn't think of a way to get the ptau from somewhere else. Keeping one on this repository is insecure right?

.github/workflows/Dockerfile
```diff
@@ -0,0 +1,30 @@
+# Install rust for Noname
+FROM rust as noname-installer
+
+# Install Noname CLI
+RUN cargo install --git https://www.github.com/zksecurity/noname
```
katat, line 5: The CLI is better to build from the current branch, otherwise this doesn't test against the latest code.

I think you can add a COPY instruction to include the noname repo into the docker image, so the CLI can be built and installed from source code.

katat (approved):
Wow, it works! Thank you so much!

Issue #132, opened 2024-06-20
support iterators by default


mimoo:
we should support iterating through an array without having to index through it. We should also support zipping arrays together. For example, in the `iterator.no` example we have:

```diff
struct Room {
    holes: Field,
}

fn Room.windows(self) -> Field {
    // a door doesn't count as a window
    return self.holes - 1;
}

struct House {
    rooms: [Room; 2],
}

fn House.room(self, const idx: Field) -> Room {
    return self.rooms[idx];
}

fn House.windows(house: House) -> Field {
    let mut windows = 0;
+    for room in house.rooms {
-    for room_idx in 0..2 {
        let room = house.room(room_idx);
        // ideally: windows +=
        windows = windows + room.windows();
    }

    return windows;
}

fn main(pub bedroom_holes: Field) -> Field {
    let bedroom = Room {
        holes: bedroom_holes,
    };

    let livingroom = Room {
        holes: 4,
    };

    let house = House { rooms: [bedroom, livingroom] };

    return House.windows(house);
}
```

PR #169, opened 2024-08-27
Add support for iterators

gio54321:
This PR aims at addressing issue #132 by introducing the possibility to iterate over arrays.
This makes it possible to write a circuit like this
```rust
fn zip(a1: [Field; 3], a2: [Field; 3]) -> [[Field; 2]; 3] {
    let mut result = [[a1[0], a1[0]], [a1[0], a1[0]], [a1[0], a1[0]]];
    for index in 0..3 {
    }
    return result;
}

fn main(pub arr: [Field; 3]) -> Field {
    let expected = [1, 2, 3];
    for pair in zip(arr, expected) {
        assert_eq(pair[0], pair[1]);
    }

    return arr[0];
}
```

This also improves the `iterate.no` test as mentioned in the issue
```rust
fn House.windows(house: House) -> Field {
    let mut windows_count = 0;
    for room in house.rooms {
        windows_count = windows_count + room.windows();
    }

    return windows_count;
}
```


src/parser/types.rs
```diff
@@ -947,80 +1013,45 @@ impl Stmt {
...
-                };
-
-                // for i in 0..5 { ... }
-                //           ^^
-                tokens.bump_expected(ctx, TokenKind::DoubleDot)?;
-
-                // for i in 0..5 { ... }
-                //             ^
-                let (end, end_span) = match tokens.bump(ctx) {
+                // here we decide if this for loop is over a range or is an iterator,
```
katat, line 1033: This is not always the case once the generic feature is introduced. For example:

```rust
fn range(const START: Field, const END: Field) {
    for ii in START..END { // START / END are const generic variables, which are not in numeric form in the syntax.
        ...
    }
}
```

So I think the safer way to detect whether it is range or iterator can be based on the presence of `..`, which indicates an range.

src/type_checker/checker.rs
```diff
@@ -610,6 +610,34 @@ impl<B: Backend> TypeChecker<B> {
...
+                body,
+            } => {
+                // enter a new scope
+                typed_fn_env.nest();
+
+                // make sure that the iterator expression is an iterator,
+                // for now this means that the expression should have type `Array`
+                let iterator_typ = self.compute_type(iterator, typed_fn_env)?;
+                let iterator_typ =
+                    iterator_typ.expect("Could not compute type of iterator (TODO: better error)");
```
katat, line 625: this can be an error thrown with a span, so it can better inform the users where the error is originated in the code.

src/type_checker/checker.rs
```diff
@@ -610,6 +610,34 @@ impl<B: Backend> TypeChecker<B> {
...
+                    typed_fn_env
+                        .store_type(var.value.clone(), TypeInfo::new(*element_type, var.span))?;
+
+                    // check block
+                    self.check_block(typed_fn_env, body, None)?;
+
+                    // exit the scope
+                    typed_fn_env.pop();
+                } else {
+                    panic!("Iterator should be an array (TODO: better error)");
```
katat, line 638: same ^

src/circuit_writer/writer.rs
```diff
@@ -5,6 +5,7 @@ use kimchi::circuits::wires::Wire;
 use num_bigint::BigUint;
 use num_traits::Num as _;
 use serde::{Deserialize, Serialize};
+use toml::value::Array;
```
katat, line 8: seems unused

katat (commented):
Looks great!

Since this is also highly related to the const generic feature , so I think it'd better to be based on the generic PR branch https://github.com/zksecurity/noname/pull/147.

src/circuit_writer/writer.rs
```diff
@@ -175,6 +174,47 @@ impl<B: Backend> CircuitWriter<B> {
                     fn_env.pop();
                 }
             }
+            StmtKind::IteratorLoop {
+                var,
+                iterator,
+                body,
+            } => {
+                let iterator_expr = self
```
katat, line 182: maybe it is better to call it as `iterator_var`, since it is of `VarOrRef` after the expression has been computed.

src/mast/mod.rs
```diff
@@ -954,6 +954,41 @@ pub fn monomorphize_stmt<B: Backend>(
...
+            let iterator_mono = monomorphize_expr(ctx, iterator, mono_fn_env)?;
+            let typ = iterator_mono.typ.as_ref().expect("expected a type");
+            let array_element_type = match typ {
+                TyKind::Array(t, _) => t,
+                _ => panic!("expected an array"),
+            };
+
+            mono_fn_env.store_type(
+                &var.value,
+                &MTypeInfo::new(&array_element_type, var.span, None),
```
katat, line 974: unnecessary & prefix

src/type_checker/checker.rs
```diff
@@ -654,6 +654,41 @@ impl<B: Backend> TypeChecker<B> {
...
+                let iterator_typ = self.compute_type(iterator, typed_fn_env)?;
+                let iterator_typ =
+                    iterator_typ.expect("Could not compute type of iterator (TODO: better error)");
+
+                // the type of the variable is the type of the items of the iterator
+                let element_type = match iterator_typ.typ {
+                    TyKind::Array(element_type, _len) => *element_type,
+                    TyKind::GenericSizedArray(element_type, _size) => *element_type,
+                    _ => {
+                        return Err(self.error(
```
katat, line 676: maybe add a negative test (`negative_tests.rs`) to cover this scenario, this would help remind this type check should be covered in the tast phase if it happens to refactor at later stage :D

katat (commented):
Nice. There are just a couple of minor issues to fix. The main one is to add the new iterator example to tests. Maybe rename the `iterators.no` to `generic_iterator.no` as it involves generic and helps differentiate from the existing `iterator.no`.

katat (approved):

katat (approved):
Awesome work! 

This feature is a great fit with the generic arrays. The generic feature enable working with varying length of arrays, while this iterator feature simplifies the use of these generic array and make it practical.

src/parser/types.rs
```diff
@@ -1258,9 +1264,53 @@ pub enum StmtKind {
         range: Range,
         body: Vec<Stmt>,
     },
+
+    // `for item in vec { <body> }`
+    IteratorLoop {
+        var: Ident,
+        iterator: Box<Expr>,
+        body: Vec<Stmt>,
+    },
```
mimoo, line 1273: late to the party, but wouldn't it be cleaner (and perhaps safer since forloops have annoying logic for generics) to have a common structure for them?

```rust
ForLoop {
  var: Ident,
  range: RangeOrIterator,
  body: Vec<Stmt>,
}
```

mimoo: cf https://github.com/zksecurity/noname/issues/177

Issue #49, opened 2024-04-23
Check the test coverage


katat:
- Do some test coverage checks on the constraint
- Each frontend constraint functions in bool/field/builtin should be covered
- A test should check the asm (this serves like a snapshot for reviewing changes)
- A test should pass the witness generation (atm, this is done by generating proof)

katat:
relevant: https://github.com/zksecurity/noname/issues/47

PR #174, opened 2024-08-31
Added test coverage checks on the constraint

ronantakizawa:
Added test coverage checks on the constraint. Resolves issue #49.

PR #179, opened 2024-09-03
Added test coverage checks on the constraint

ronantakizawa:
Added test coverage checks on the constraint. Resolves issue #49. Continuing from PR #174 (Made a separate PR since it was easier to add the new tests from a new branch into branch `rfc-0-impl` instead of the old branch)

### DIFF
--- a/src/negative_tests.rs
+++ b/src/negative_tests.rs
@@ -1,45 +1,12 @@
 use crate::{
-    backends::{
-        kimchi::KimchiVesta,
-        r1cs::{R1csBn254Field, R1CS},
-        Backend,
-    },
+    backends::r1cs::{R1csBn254Field, R1CS},
     circuit_writer::CircuitWriter,
-    compiler::{compile, generate_witness, typecheck_next_file_inner, Sources},
+    compiler::{typecheck_next_file_inner, Sources},
     error::{ErrorKind, Result},
-    inputs::parse_inputs,
     mast::Mast,
     type_checker::TypeChecker,
     witness::CompiledCircuit,
 };
-
-fn ast_pass(code: &str) -> Result<()> {
-    let mut source = Sources::new();
-    typecheck_next_file_inner(
-        &mut TypeChecker::<R1CS<R1csBn254Field>>::new(),
-        None,
-        &mut source,
-        "example.no".to_string(),
-        code.to_string(),
-        0,
-    )
-    .map(|_| ())
-}
-
-fn nast_pass(code: &str) -> (Result<usize>, TypeChecker<R1CS<R1csBn254Field>>, Sources) {
-    let mut source = Sources::new();
-    let mut nast = TypeChecker::<R1CS<R1csBn254Field>>::new();
-    let res = typecheck_next_file_inner(
-        &mut nast,
-        None,
-        &mut source,
-        "example.no".to_string(),
-        code.to_string(),
-        0,
-    );
-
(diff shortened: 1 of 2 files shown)


src/negative_tests.rs
```diff
@@ -1,13 +1,46 @@
...
+        None,
+        &mut source,
+        "example.no".to_string(),
+        code.to_string(),
+        0,
+    )
+    .map(|_| ())
+}
+
+fn nast_pass(code: &str) -> (Result<usize>, TypeChecker<R1CS<R1csBn254Field>>, Sources) {
```
katat, line 29: this is same as the `tast_pass`

src/negative_tests.rs
```diff
@@ -1,13 +1,46 @@
...
-    compiler::{typecheck_next_file_inner, Sources},
+    compiler::{compile, generate_witness, typecheck_next_file_inner, Sources},
     error::{ErrorKind, Result},
+    inputs::parse_inputs,
     mast::Mast,
     type_checker::TypeChecker,
     witness::CompiledCircuit,
 };
 
+fn ast_pass(code: &str) -> Result<()> {
```
katat, line 16: this is same as the `tast_pass`

src/negative_tests.rs
```diff
@@ -321,3 +385,99 @@ fn test_iterator_variable_redefinition() {
         ErrorKind::DuplicateDefinition(..)
     ));
 }
+
+#[test]
+fn test_boolean_and_fail() {
```
katat, line 390: yeah, this is a good addition for testing a negative tast case.

src/negative_tests.rs
```diff
@@ -321,3 +385,99 @@ fn test_iterator_variable_redefinition() {
...
+        return yy + true;
+    }
+    "#;
+
+    let res = tast_pass(code).0;
+    assert!(matches!(res.unwrap_err().kind, ErrorKind::MismatchType(..)));
+}
+
+#[test]
+fn test_asm_snapshot_mismatch_fail() {
```
katat, line 440: The asm will be compared in the `examples.rs`. It is more a higher level testing category.

The test cases here are more about testing the expected exceptions, at lower level.

src/negative_tests.rs
```diff
@@ -321,3 +385,99 @@ fn test_iterator_variable_redefinition() {
...
+
+    let expected_asm_output = "expected output that you expect to match";
+    assert!(
+        asm_output.trim() != expected_asm_output.trim(),
+        "Expected an ASM mismatch error, but no mismatch was detected."
+    );
+}
+
+#[test]
+fn test_invalid_witness_generation_fail() {
```
katat, line 464: similarly, the witness generation should be covered in the `examples.rs`.

katat (commented):
I think these tests are getting more relevant. Some of the additions indeed contribute to the coverage at the relevant level.

I think maybe we just keep this `negative_test.rs` cover the negative tests for the tast / mast / synthesizer phases.

For the other phases, like parsing AST and NAST, we would just add the coverage in their existing locations.

Existing coverage for AST: https://github.com/zksecurity/noname/blob/1a99e8353db4b229eeff72bd9efa22f486c27d28/src/parser/mod.rs#L193

Existing coverage for NAST: https://github.com/zksecurity/noname/blob/9a06998dc600d1a03f18c4f8fef11696ac400f8b/src/name_resolution/mod.rs#L77

PR #183, opened 2024-09-11
fix: properly propagate constant value to assigned var in mast phase

katat:
This PR allows propagating constant values in assignment statement.

Sometimes we want to propagate a constant value from the right hand side to a variable in assignment, eg:

```rust
let var = CST + 1 // this PR fix this: constant value can't be propagated to var
```

Meanwhile, we need to ensure it won't propagate to the constant value and treat the mutable variable as a constant. Otherwise, in some cases, the mutable variable will mistakenly folded as a constant. eg:

```rust
fn accumulate_mut(const LEN: Field) -> Field {
    // shouldn't fold mutable variables as constants
    let mut zz = LEN; // assign a constant value
    for ii in 0..3 {
        zz = zz + zz; // if zz is treated as a constant, mutable variable zz will be mistakenly folded as a constant
    }
    return zz;
}
```

The reason it needs to fold the constant values during mast phase is because during initializing a generic function, the constants need to be folded to resolve the concrete value. eg:
```rust
fn last(arr: [[Field; NN]; 3]) -> Field {
    let mut newarr = [0; NN * 3]; // <-- NN * 3 has to be folded
    ...
```


katat:
On a related topic, I am wondering if we should allow either const or non-const field for condition branches, eg:
```rust
fn from_bits(bits: [Bool; LEN]) -> Field {
    let mut lc1 = 0;
    let mut e2 = 1; // non-constant
    let zero = 0;  // constant

    for index in 0..LEN {
        // the current type check throws error from the if then else condition check
        // branches should allow either const or non const field?
        lc1 = lc1 + if bits[index] {e2} else {zero};
        e2 = e2 + e2;
    }
    return lc1;
}
```

mimoo (approved):
As discussed I think this should be a temporary solution and we should create an issue/discussion on how to make this kind of code work :o

katat:
wrote up an issue to better address this problem: https://github.com/zksecurity/noname/issues/188

katat:
I reverted the changes, as it is non issue any more because they are better fixed in another PR. See the comment: https://github.com/zksecurity/noname/issues/188#issuecomment-2365134069

Issue #176, opened 2024-09-02
Merge forloop structure into one


katat:
Currently there are two structures for the `ForLoop` statement:

```rust
    // `for var in 0..10 { <body> }`
    ForLoop {
        var: Ident,
        range: Range,
        body: Vec<Stmt>,
    },

    // `for item in vec { <body> }`
    IteratorLoop {
        var: Ident,
        iterator: Box<Expr>,
        body: Vec<Stmt>,
    },
```

It would be nicer and safer to merge them into one (https://github.com/zksecurity/noname/pull/169#discussion_r1741215859)

```rust
ForLoop {
  var: Ident,
  range: RangeOrIterator,
  body: Vec<Stmt>,
}
```

PR #186, opened 2024-09-19
Refactor for loops and iterators in the same statement structure

gio54321:
This PR aims at addressing issue #176 by refactoring the `ForLoop` and `IteratorLoop` structures into a single structure

```rust
ForLoop {
    var: Ident,
    argument: ForLoopArgument,
    body: Vec<Stmt>,
}
```

using the `ForLoopArgument` enum as the argument type.

```rust
pub enum ForLoopArgument {
    Range(Range),
    Iterator(Box<Expr>),
}
```

Indeed this would have been much cleaner to implement iterators in this way from the beginning! I tried to merge as much as possible the code in the `ForLoop` and `IteratorLoop` cases throughout the compilation flow, but sometimes it ended up as a big `match` statement on the argument (e.g., in `circuit_writer.rs`).
Maybe in the future this could be refactored even further introducing a generic iterator trait in the codebase, having ranges and vectors as implementations.

### DIFF
--- a/src/circuit_writer/writer.rs
+++ b/src/circuit_writer/writer.rs
@@ -13,7 +13,7 @@ use crate::{
     error::{ErrorKind, Result},
     imports::FnKind,
     parser::{
-        types::{FunctionDef, Stmt, StmtKind, TyKind},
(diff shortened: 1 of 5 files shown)


katat (approved):
Looks good to me!

Issue #185, opened 2024-09-19
Functions are not always monomorphized


gio54321:
I was trying to write this simple circuit
```rust
fn sum_arr(arr1: [Field; 3], arr2: [Field; 3]) -> [Field; 3] {
    let mut result = [0; 3];
    for idx in 0..3 {
        result[idx] = arr1[idx] + arr2[idx];
    }
    return result;
}

fn main(pub arr: [Field; 3]) -> [Field; 3] {
    let result = sum_arr(arr, arr);
    return result;
}
```

But, trying to compile it, gives back a panic in [circuit writer](https://github.com/zksecurity/noname/blob/main/src/circuit_writer/writer.rs#L700-L711)

```
thread 'main' panicked at src/circuit_writer/writer.rs:713:26:
expected array
```

After a bit of debugging, the problem is that the lhs of the assignment 
```
result[idx] = arr1[idx] + arr2[idx];
```
still has type `GenericSizedArray(Field { constant: true }, Concrete(3))`, which should not happen since we are in the circuit writer component.

I think the main issue here is that the body of the function `sum_arr` is not monomorphized, since functions are monomorphized [only if needed](https://github.com/zksecurity/noname/blob/main/src/mast/mod.rs#L413-L436). Debugging the code,  `fn_info.sig().require_monomorphization()` returns False (which makes sense as the function signature does not contain any generic-sized array), so the function body AST is left as is and passed to the circuit writer component. However the AST contains the `GenericSizedArray` objects because of the generic sized array `result` declared inside the function.

katat:
Great investigation!

Yes, indeed this is a bug. I think the solution can be changing the `GenericSizedArray` to the concrete type `Array` when propagating the type during type checking phase.

```rust
ExprKind::RepeatedArrayInit { item, size } => {
    let item_node = self
        .compute_type(item, typed_fn_env)?
        .expect("expected a typed item");

    let size_node = self
        .compute_type(size, typed_fn_env)?
        .expect("expected a typed size");

    if is_numeric(&size_node.typ) {
        // use generic array as the size node might include generic parameters or constant vars
        let res = ExprTyInfo::new_anon(TyKind::GenericSizedArray(
            Box::new(item_node.typ),
            Symbolic::parse_expr(size)?,
        ));
        Some(res)
    } else {
        return Err(self.error(ErrorKind::InvalidArraySize, expr.span));
    }
}
```

[...]

PR #187, opened 2024-09-20
Fix function instantiation

katat:
This PR is for https://github.com/zksecurity/noname/issues/185

### Changes
1. Instantiate all functions, no matter whether it is generic function not. This is to resolve the case that @gio54321 pointed out below. 
2. Avoid unnecessary `GenericSizedArray` type
3. Fix scope tracking for monomorphizing a function. It helps determine whether it is within a scope being monomorphized so as to decide whether to update the AST. This is not necessary if all functions are instantiated at the moment, but will be needed if decided to use a more efficient way to monomorphize the functions instead of all.

### Another potential approach
To properly avoid instantiating all functions but only the ones containing generic vars or function calls, it would need an additional function AST walker to check if it is a function needed to be instantiated, or add a flag `require_instantiation` during the AST walk in TAST phase.

Then it can properly determine whether to instantiate a function, instead of simply relying on checking if it contains generic variables in the function signature.


gio54321:
This fix was quick! :smile: 
I don't think however that this approach solves the issue in the general case. Take for example this circuit, which adds an additional call to a generic function
```rust
fn zip(a1: [Field; LEN], a2: [Field; LEN]) -> [[Field; 2]; LEN] {
    let mut result = [[0; 2]; LEN];
    for index in 0..LEN {
    }
    return result;
}

fn sum_arr(arr1: [Field; 3], arr2: [Field; 3]) -> [Field; 3] {
    let mut result = [0; 3];
    let pairs = zip(arr1, arr2);
    for idx in 0..3 {
        let item = pairs[idx];
        result[idx] = item[0] + item[1];
    }
    return result;
}

fn main(pub arr: [Field; 3]) -> [Field; 3] {
    let result = sum_arr(arr, arr);
    return result;
}
```

The function `zip` requires monomrphization and it is instantiated in `sum_arr` with `LEN=3`, but the body of `sum_arr` is not monomorphized. In my head, the need for monomorphization cannot be derived only from the signature but rather from the call graph, i.e., it is kind of a transitive property

A possible approach would be to always monomorphize the body of the functions recursively, and then put in scope the monomorphized symbol (I mean the `zip#LEN=3` symbol) only if the function signature is generic.

katat:
Cool, thanks for taking a look :)

> A possible approach would be to always monomorphize the body of the functions recursively, and then put in scope the monomorphized symbol (I mean the zip#LEN=3 symbol) only if the function signature is generic.

I think it already works this way. Even though the body of `sum_arr` is not monomorphized, the function `zip` should be monomorphized. It should work recursively already. But maybe I miss something :o

The current rule for monomorphization is:
- walk through the AST nodes and check the function calls
- if the function call contain generic variables, such as `fn zip(a1: [Field; LEN], a2: [Field; LEN]) -> [[Field; 2]; LEN]`, it will initialize it, replace the `FnCall` node' fully qualified with the monomorphized function name something like `zip#LEN=3`

Then in synthesizer phase (circuit writer), the replaced `FnCall` node should call look up the monomorphized function AST instead of the origin one.

katat:
but yeah, your example doesn't compile. I am taking a look.

katat:
Indeed, this is an issue as you described above. This is a great spot! @gio54321 

We used to instantiate all the functions, including the non generic ones. But we tried to only instantiate the generic functions while keeping the non generic function as it is in order to minimize the affected scopes after monomorphization phase. That is why the use of `fn_info.sig().require_monomorphization()` to determine it needs to instantiate the function.

The problem this code doesn't work as I described above is because it only keeps walking into the generic functions when it handles `FnCall` node.

Therefore, I think updating the handling of `FnCall` to keep it walk through all the function calls besides the generic functions should fix the issue.

mimoo (approved):
LGTM

PR #193, opened 2024-09-24
feat: log builtin

katat:
This is a logging feature to help debugging.

Usage:
```rust
let sum = var1 + var2;
log(sum); 
```

It should print something like:
```
---log span: Span { filename_id: 1, start: 147, len: 7 }---
typ: Some(Field { constant: false })
mutable: false
cvar: v_2 + v_1
```

Because it is a builtin function, so it only accesses to the info at compilation time.

It is quite common to check the values behind these variables. So we could record the cell var index at the backend, and print out the relevant witness values when generating witnesses.

### DIFF
--- a/src/backends/r1cs/mod.rs
+++ b/src/backends/r1cs/mod.rs
@@ -39,12 +39,38 @@ impl CellVar {
 
 impl<F: BackendField> BackendVar for LinearCombination<F> {}
 
+impl<F: BackendField> std::fmt::Debug for LinearCombination<F> {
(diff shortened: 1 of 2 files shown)


mimoo:
so two things:

1. I think at some point we'll want a real format syntax like in rust (e.g. `log("{sum}")`)
2. we should distinguish debugging runtime value and type information, here it looks more like a `log("{}", typeof(sum))` (or perhaps we could have a separate `log_type` function)

mimoo (approved):

Issue #194, opened 2024-09-26
Support const attribute to custom type's field


katat:
To better type check a const field from a custom type, it would be better to support:

```rust
Thing {
  xx: Field,
  yy: const Field
}
```

mimoo:
^ as discussed privately I don't think this is a good approach (I need to copy/paste the arguments I had here)

PR #195, opened 2024-09-26
allow use custom's const field to declare symbolic size

katat:
This makes it able to declare methods to return array using the struct's field as its symbolic size.

For example:

```rust
fn Thing.to_array(self) -> [Field; self.cst] {
    return [1; self.cst];
}
```

## Changes
1. Refactored the constant propagation to allow propagating multiple constants, such as array and custom type, in addition to single constant.
2. Allow propagating constant values from native function/method.

I also create [an issue](https://github.com/zksecurity/noname/issues/194) to propose supporting `const` attribute for custom fields for better type checking. So it can type check `self.cst` for [Field; self.cst] where the symbolic size has to be constant.


mimoo:
as discussed privately I don't think we should go this route :o

katat:
agreed, closing this.

PR #198, opened 2024-10-09
Fix: avoid node overriding and re-instantiate a same function

katat:
Changes: 

1. Since every functions need to be instantiated, the `ctx.generic_func_scope` is not necessary for determining whether it needs to generate new node id. Thus, `ctx.generic_func_scope` is removed to avoid unexpected behavior in terms of handling the node ids at mast phase.

2. A fix to avoid re-instantiate a same function.

### DIFF
--- a/src/mast/mod.rs
+++ b/src/mast/mod.rs
@@ -152,10 +152,10 @@ where
 {
     tast: TypeChecker<B>,
     generic_func_scope: Option<usize>,
-    // fully qualified function name
-    functions_to_delete: Vec<FullyQualified>,
-    // fully qualified struct name, method name
-    methods_to_delete: Vec<(FullyQualified, String)>,
+    // new fully qualified function name as the key, old fully qualified function name as the value
+    functions_instantiated: HashMap<FullyQualified, FullyQualified>,
+    // new method name as the key, old method name as the value
+    methods_instantiated: HashMap<(FullyQualified, String), String>,
 }
 
 impl<B: Backend> MastCtx<B> {
@@ -163,8 +163,8 @@ impl<B: Backend> MastCtx<B> {
         Self {
             tast,
             generic_func_scope: Some(0),
-            functions_to_delete: vec![],
-            methods_to_delete: vec![],
+            functions_instantiated: HashMap::new(),
+            methods_instantiated: HashMap::new(),
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/mast/mod.rs
```diff
@@ -417,26 +421,45 @@ fn monomorphize_expr<B: Backend>(
...
-            let mexpr = Expr {
-                kind: ExprKind::FnCall {
-                    module: module.clone(),
-                    fn_name: fn_name_mono.clone(),
-                    args: args_mono,
-                },
-                ..expr.clone()
-            };
+            // check if this function is already monomorphized
+            if ctx.functions_instantiated.contains_key(&old_qualified) {
```
gio54321, line 427: Shouldn't this be checked to contain the new signature as a key? I think in the code below the `zip` function is monomorphized two times, even if the LEN parameter is the same. Though, maybe I'm missing the point of the PR :smile: 
```rust
fn zip(a1: [Field; LEN], a2: [Field; LEN]) -> [[Field; 2]; LEN] {
    let mut result = [[0; 2]; LEN];
    for index in 0..LEN {
    }
    return result;
}

fn main(pub arr: [Field; 3]) -> Field {
    let expected = [1, 2, 3];
    for pair in zip(arr, expected) {
        assert_eq(pair[0], pair[1]);
    }
    for pair in zip(arr, expected) {
        assert_eq(pair[0], pair[1]);
    }

    return arr[0];
}
```

src/mast/mod.rs
```diff
@@ -490,25 +513,46 @@ fn monomorphize_expr<B: Backend>(
                 args_mono.push(expr_mono.expr);
             }
 
-            // monomorphize the function call
-            let (fn_info_mono, typ) = instantiate_fn_call(ctx, fn_info, &observed, expr.span)?;
+            // check if this function is already monomorphized
+            if ctx
+                .methods_instantiated
+                .contains_key(&(struct_qualified.clone(), method_name.value.clone()))
```
gio54321, line 519: Same as before

gio54321 (commented):
Looks good!
Left a couple of comments as it's not clear to me the usage of the mapping

src/mast/mod.rs
katat: Oh yeah, indeed. It should always check the monomorphized name.

I have just pushed a fix that also includes a refactor to better encapsulate the resolutions for the generic values and types. This refactor makes it always resolves the types so as to get the monomorphized qualified name, without having to re-instantiate a function.

gio54321 (approved):
LGTM now! Also very nice refactoring :smiley:

PR #199, opened 2024-10-09
Support hint function declaration to bind with builtin function

katat:
This is the first step to support hint function, by making use the builtin functions as hint functions.

Although it is not yet fully supporting the native hint function, this PR makes it able to type check the `unsafe/hint` attribute.

Usage:

Declares a hint function in a `.no` file to wire up with a builtin function under the same full qualified name.
```rust
hint fn calc(val: Field) -> Field;
```

It requires `unsafe` attribute when calling a hint function, otherwise the type checker will throw error. So the hint function caller has to use `unsafe` to acknowledge they are calling a hint function.
```rust
fn main(pub xx: Field) {
    let yy = unsafe calc(xx);
}
```

This can make the type checking ready for hint calculations before the native hint function support. 
Once the hint function support is ready, we can just fill the hint function body in the native code and deprecate the hint builtin functions, while keeping existing hint function callers intact.


src/parser/mod.rs
```diff
@@ -200,6 +220,7 @@ mod tests {
         let code = r#"main(pub public_input: [Fel; 3], private_input: [Fel; 3]) -> [Fel; 3] { return public_input; }"#;
         let tokens = &mut Token::parse(0, code).unwrap();
         let ctx = &mut ParserCtx::default();
+        let is_hint = false;
```
mimoo, line 223: ?

src/type_checker/checker.rs
```diff
@@ -141,6 +147,16 @@ impl<B: Backend> TypeChecker<B> {
...
 
+                // check if the function is a hint
+                if fn_info.is_hint && !unsafe_attr {
+                    return Err(self.error(ErrorKind::ExpectedUnsafeAttribute, expr.span));
+                }
+
+                // unsafe attribute should only be used on hints
+                if !fn_info.is_hint && *unsafe_attr {
+                    return Err(self.error(ErrorKind::UnexpectedUnsafeAttribute, expr.span));
+                }
```
mimoo, line 158: maybe more concise:

```rust
match (fn_info.is_hint, unsafe_attr) {
  (true, true) | (false, false) => (),
  _ => return Err(self.error(ErrorKind::ExpectedUnsafeAttribute, expr.span));
  }
```

src/parser/types.rs
```diff
@@ -1161,7 +1162,33 @@ impl FunctionDef {
...
         }
 
-        let func = Self { sig, body, span };
+        let func = Self { sig, body, span, is_hint: false };
+
+        Ok(func)
+    }
+
+    /// Parse a hint function signature
+    pub fn parse_hint(ctx: &mut ParserCtx, tokens: &mut Tokens) -> Result<Self> {
```
mimoo, line 1164: do you think it'd make sense to merge this with the parse function above?

mimoo (approved):
LGTM, I'll have another pass but let's not wait to merge it!

katat:
actually this PR is already part of https://github.com/zksecurity/noname/pull/201 xD

src/parser/types.rs
katat: This new parser function is for hint function without a body.
So I think once we support the native hints, we can just remove this function, while the native hint function can get parsed using parser function above without changes.

src/type_checker/checker.rs
katat: Actually there are two error types to handle: https://github.com/zksecurity/noname/blob/18825a3a4a53572bd038c679e823ffc5d2e17ff9/src/negative_tests.rs#L476-L500

PR #204, opened 2024-10-17
Fix: Allow struct fields to propagate constants

katat:
This is a fix to propagate the constant values from struct fields also.

This fix allows the struct fields to be passed as an constant argument for a generic function. eg:

```rust
struct Uint8 {
    inner: Field,
    bit_len: Field,
}

fn Uint8.new(val: Field) -> Uint8 {
    let bit_len = 8;

    // range check
    let ignore_ = bits::to_bits(bit_len, val);

    return Uint8 {
        inner: val,
        bit_len: bit_len
    };
}

fn Uint8.less_than(self, rhs: Uint8) -> Bool {
    // this won't work without this PR fix
    return comparator::less_than(self.bit_len, self.inner, rhs.inner);
}
```

### DIFF
--- a/src/mast/mod.rs
+++ b/src/mast/mod.rs
@@ -9,7 +9,8 @@ use crate::{
     imports::FnKind,
     parser::{
         types::{
-            FnSig, ForLoopArgument, GenericParameters, Range, Stmt, StmtKind, Symbolic, Ty, TyKind,
+            FnSig, ForLoopArgument, GenericParameters, Ident, Range, Stmt, StmtKind, Symbolic, Ty,
+            TyKind,
         },
         CustomType, Expr, ExprKind, FunctionDef, Op2,
     },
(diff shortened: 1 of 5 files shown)


gio54321 (approved):
Looks good to me, I did not find any issues in the implementation, this is definitely a fix worth merging in the stdlib PR :smile:

I'm wondering, defining `Uint8` using a structure allows an user to modify the bit length, which maybe is not necessarily what we want. Perhaps it is better to define it using a constant `BIT_LEN` so that we ensure that it is fixed in the monomorphization phase.
Afaiu, if the user using `Uint8` do modify the bit length with another constant, the monomorphization would still pass with the new bit length, but maybe I'm wrong.

katat:
Indeed it is potentially an issue when it comes to the fact that the users can modify the field, such as bit_len, which supposes to be a constant and can't be changed once the struct is declared. There is an ongoing discussion on this issue: https://github.com/zksecurity/noname/issues/194

PR #206, opened 2024-10-18
Add u16/32/64

katat:
- Add builtin types: u16/32/64

- Increate test coverage for these newly added stdlib


src/stdlib/native/int.no
```diff
@@ -18,10 +19,90 @@ fn Uint8.new(val: Field) -> Uint8 {
     };
 }
 
+// u16
+struct Uint16 {
+    inner: Field,
+    bit_len: Field,
```
mimoo, line 25: as I said elsewhere, I don't think we should carry that in the struct and we should just make sure we use `16` as argument in the right places

src/stdlib/native/int.no
```diff
@@ -18,10 +19,90 @@ fn Uint8.new(val: Field) -> Uint8 {
...
+struct Uint32 {
+    inner: Field,
+    bit_len: Field,
+}
+
+fn Uint32.new(val: Field) -> Uint32 {
+    let bit_len = 32;
+
+    // range check
+    let ignore_ = bits::to_bits(bit_len, val);
```
mimoo, line 50: usually the syntax is `let _ =` or `let _bits = `

src/stdlib/native/int.no
```diff
@@ -18,10 +19,90 @@ fn Uint8.new(val: Field) -> Uint8 {
...
+fn Uint32.less_eq_than(self, rhs: Uint32) -> Bool {
+    return comparator::less_eq_than(self.bit_len, self.inner, rhs.inner);
+}
+
+fn Uint64.less_than(self, rhs: Uint64) -> Bool {
+    return comparator::less_than(self.bit_len, self.inner, rhs.inner);
+}
+
+fn Uint64.less_eq_than(self, rhs: Uint64) -> Bool {
+    return comparator::less_eq_than(self.bit_len, self.inner, rhs.inner);
```
mimoo, line 107: we should also implement `add`, `sub`, etc.
this is when having a `bit_len` field would be interesting... because theoretically we could track how many bits max the result of an operation is, and only trigger a range check when we might be above 8 bits for u8. But in the absence of that we can simply perform a range check after every operation

mimoo (approved):
LGTM

src/stdlib/native/int.no
katat: the current rule doesn't allow starting with `_`

src/stdlib/native/int.no
katat: removed

src/stdlib/native/int.no
katat: To take it further, maybe we can make it support builtin operations, such as + -, by mapping these operations to these functions. https://github.com/zksecurity/noname/issues/214

PR #208, opened 2024-10-23
Add MIMC stdlib

katat:
- Fix: Default constants to bigint type in order to allow larger literals in native code, such as the constant values in mimc.
- Added MIMC stdlib(for a single input), but it requires the support of generic calls within forloop in order to support `multi_mimc`: https://github.com/zksecurity/noname/pull/189


src/mast/mod.rs
```diff
@@ -224,7 +224,10 @@ impl FnSig {
             TyKind::Array(ty, size) => TyKind::Array(Box::new(self.resolve_type(ty, ctx)), *size),
             TyKind::GenericSizedArray(ty, sym) => {
                 let val = sym.eval(&self.generics, &ctx.tast);
-                TyKind::Array(Box::new(self.resolve_type(ty, ctx)), val)
+                TyKind::Array(
+                    Box::new(self.resolve_type(ty, ctx)),
+                    val.to_u32().expect("array size exceeded u32"),
```
mimoo, line 229: use that `UnexpectedError` from https://github.com/zksecurity/noname/pull/209 :o

mimoo (approved):
LGTM!

PR #209, opened 2024-10-23
"PR Init (error msgs and handling): sample changes for feedback"

### LINKED ISSUE #26: Replace all panics with cleaner error
We should replace all panics with an `ErrorKind::UnexpectedError(&'static str)` or something like that. So that a user can paste the error in the issues here if they run into one!

This error could print a message like:

> You've reached an unknown error, please copy/paste the rest of this error on https://github.com/zksecurity/noname/issues/new?title=unknown%20error
> [the string]

see https://github.com/zksecurity/noname/pull/209 for further guidance on how to convert panics to errors

in general we should not replace panics with specific errors as we don't know if we'll ever reach them. Specific errors are good if we know the user can reach them by writing bad code. 

In general the flow should be this:

* any panic in code should return an UnexpectedError to the user
* if the user reports such an error, we know that we either have a bug (we assumed that this code should never be reached, but it was reached) or we have an opportunity to write a better error for that specific error

crypdoughdoteth:
Closes #26 - removes panics in favor of clear error messages and improves error handling in some places

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -41,15 +41,12 @@ pub enum ErrorKind {
         "Unexpected error: {0}. Please report this error on https://github.com/mimoo/noname/issues"
     )]
     UnexpectedError(&'static str),
-
     #[error("variable is not mutable. You must set the `mut` keyword to make it mutable")]
     AssignmentToImmutableVariable,
-
     #[error(
         "the dependency `{0}` does not appear to be listed in your manifest file `Noname.toml`"
     )]
     UnknownDependency(String),
-
     #[error("the function `{1}` does not exist in the module `{0}`")]
     UnknownExternalFn(String, String),
 
--- a/src/inputs.rs
+++ b/src/inputs.rs
@@ -34,6 +34,14 @@ pub enum ParsingError {
 
     #[error("mismatch between expected argument format ({0}), and given argument in JSON (`{1}`)")]
     MismatchJsonArgument(TyKind, serde_json::Value),
+    #[error("Incorrect array size: Expected {0}, Got {1})")]
+    ArraySizeMismatch(usize, usize),
+    #[error("Incorrect struct fields count. {0} expected {1}, Got {2})")]
+    StructFieldCountMismatch(String, usize, usize),
+    #[error("Failed to locate struct name `{0}` given as input")]
+    MissingStructIdent(String),
+    #[error("Failed to locate struct field name `{0}` in JSON input")]
+    MissingStructFieldIdent(String),
 }
 
 //
@@ -93,7 +101,10 @@ impl<B: Backend> CompiledCircuit<B> {
 
             (TyKind::Array(el_typ, size), Value::Array(values)) => {
                 if values.len() != (*size as usize) {
-                    panic!("wrong size of array");
+                    Err(ParsingError::ArraySizeMismatch(
+                        values.len(),
+                        *size as usize,
+                    ))?
                 }
                 let mut res = vec![];
                 for value in values {
@@ -115,23 +126,24 @@ impl<B: Backend> CompiledCircuit<B> {
                 let struct_info = self
                     .circuit
                     .struct_info(&qualified)
-                    .expect("compiler bug: couldn't find struct given as input");
+                    .ok_or_else(|| ParsingError::MissingStructIdent(struct_name.to_string()))?;
                 let fields = &struct_info.fields;
 
                 // make sure that they're the same length
                 if fields.len() != map.len() {
-                    panic!("wrong number of fields in struct (TODO: better error)");
+                    Err(ParsingError::StructFieldCountMismatch(
+                        struct_info.name.to_string(),
+                        map.len(),
+                        fields.len(),
+                    ))?;
                 }
 
                 // parse each field
                 let mut res = vec![];
                 for (field_name, field_ty) in fields {
                     let value = map
-                    .remove(field_name)
-                    .ok_or_else(|| {
-                        format!("couldn't find field `{field_name}` in given JSON input (TODO: better error)")
-                    })
-                    .unwrap();
+                        .remove(field_name)
+                        .ok_or_else(|| ParsingError::MissingStructFieldIdent(field_name.to_string()))?;
                     let parsed = self.parse_single_input(value, field_ty)?;
                     res.extend(parsed);
                 }
--- a/src/mast/mod.rs
+++ b/src/mast/mod.rs
@@ -433,8 +433,11 @@ pub fn monomorphize<B: Backend>(tast: TypeChecker<B>) -> Result<Mast<B>> {
     let mut func_def = match &main_fn.kind {
         // `fn main() { ... }`
         FnKind::Native(function) => function.clone(),
-
-        _ => panic!("main function must be native"),
+        _ => Err(Error::new(
+            "Backend - Monomorphize",
+            ErrorKind::UnexpectedError("Main function must be native"),
+            Span::default(),
+        ))?,
     };
 
     // create a new typed fn environment to type check the function
@@ -484,7 +487,11 @@ fn monomorphize_expr<B: Backend>(
             // obtain the type of the field
             let (module, struct_name) = match lhs_mono.typ {
                 Some(TyKind::Custom { module, name }) => (module, name),
-                _ => panic!("field access must be done on a custom struct"),
+                _ => Err(Error::new(
+                    "Monomorphize Expr",
+                    ErrorKind::UnexpectedError("field access must be done on a custom struct"),
+                    expr.span,
+                ))?,
             };
 
             // get struct info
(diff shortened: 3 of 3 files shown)


crypdoughdoteth:
Quick question: in mast/mod.rs/42:13: should the constructor of `ExprMonoInfo` panic?

src/error.rs
```diff
@@ -41,15 +41,12 @@ pub enum ErrorKind {
         "Unexpected error: {0}. Please report this error on https://github.com/mimoo/noname/issues"
     )]
     UnexpectedError(&'static str),
```
mimoo, line 43: maybe a good idea to change the comment there, since we don't use it for testing anymore, but to warn users with nicer errors than panics

src/inputs.rs
```diff
@@ -93,7 +101,10 @@ impl<B: Backend> CompiledCircuit<B> {
 
             (TyKind::Array(el_typ, size), Value::Array(values)) => {
                 if values.len() != (*size as usize) {
-                    panic!("wrong size of array");
+                    Err(ParsingError::ArraySizeMismatch(
+                        values.len(),
+                        *size as usize,
+                    ))?
```
mimoo, line 107: in general we should not replace panics with specific errors as we don't know if we'll ever reach them. Specific errors are good if we know the user can reach them by writing bad code. In this case though, we clearly have a reachable error so it's a good change!

src/inputs.rs
```diff
@@ -115,23 +126,24 @@ impl<B: Backend> CompiledCircuit<B> {
                 let struct_info = self
                     .circuit
                     .struct_info(&qualified)
-                    .expect("compiler bug: couldn't find struct given as input");
+                    .ok_or_else(|| ParsingError::MissingStructIdent(struct_name.to_string()))?;
```
mimoo, line 129: that's an example of an error that would benefit from just being an `UnexpectedError` here. In general the flow should be this:

* any panic in code should return an `UnexpectedError` to the user
* if the user reports such an error, we know that we either have a bug (we assumed that this code should never be reached, but it was reached) or we have an opportunity to write a better error for that specific error

src/inputs.rs
```diff
@@ -115,23 +126,24 @@ impl<B: Backend> CompiledCircuit<B> {
...
                 let fields = &struct_info.fields;
 
                 // make sure that they're the same length
                 if fields.len() != map.len() {
-                    panic!("wrong number of fields in struct (TODO: better error)");
+                    Err(ParsingError::StructFieldCountMismatch(
+                        struct_info.name.to_string(),
+                        map.len(),
+                        fields.len(),
+                    ))?;
```
mimoo, line 138: 👍

src/inputs.rs
```diff
@@ -115,23 +126,24 @@ impl<B: Backend> CompiledCircuit<B> {
...
                 let mut res = vec![];
                 for (field_name, field_ty) in fields {
                     let value = map
-                    .remove(field_name)
-                    .ok_or_else(|| {
-                        format!("couldn't find field `{field_name}` in given JSON input (TODO: better error)")
-                    })
-                    .unwrap();
+                        .remove(field_name)
+                        .ok_or_else(|| ParsingError::MissingStructFieldIdent(field_name.to_string()))?;
```
mimoo, line 146: 👍

mimoo (approved):
This is great, thanks so much! I'll merge this one but feel free to address the comments in another PR :)

mimoo:
actually I see this is a draft, so let me know when it's good to be merged!

crypdoughdoteth:
Panics are now limited to circuit_writer.rs and some constructors

src/inputs.rs
crypdoughdoteth: I agree. I did it this way because the function returns ParsingError. I added some extra context in the message so we know it's a bug when the user encounters it

src/backends/r1cs/snarkjs.rs
```diff
@@ -39,16 +39,20 @@ struct SnarkjsLinearCombination {
...
+
+            Err(std::io::Error::new(
+                ErrorKind::Other,
+                "The first var should be preserved for constant term",
+            ))?
         }
 
-        terms
+        Ok(terms)
     }
```
mimoo, line 56: is this necessary @katat ?

src/circuit_writer/mod.rs
```diff
@@ -247,7 +247,7 @@ impl<B: Backend> CircuitWriter<B> {
         // add argument variable to the ast env
         let mutable = false; // TODO: should we add a mut keyword in arguments as well?
         let var_info = VarInfo::new(var, mutable, Some(typ.kind.clone()));
-        self.add_local_var(fn_env, name.value.clone(), var_info);
+        self.add_local_var(fn_env, name.value.clone(), var_info)?;
```
mimoo, line 250: there are more instances that need to be converted now

src/error.rs
```diff
@@ -37,12 +37,12 @@ impl Error {
 #[derive(Error, Diagnostic, Debug)]
 pub enum ErrorKind {
     /// this error is for testing. You can use it when you want to quickly see in what file and what line of code the error is.
-    #[error(
-        "Unexpected error: {0}. Please report this error on https://github.com/mimoo/noname/issues"
-    )]
+    #[error("An unexpected error occured: {0}")]
```
mimoo, line 40: oh the text was good! We should still ask people to report the error, I meant the comment above

src/stdlib/builtins.rs
```diff
@@ -51,17 +51,35 @@ fn assert_eq_fn<B: Backend>(
...
+                "constraint-generation",
+                ErrorKind::UnexpectedError("No type info for rhs of assertion"),
+                span,
+            )
+        })?;
+
+        Err(Error::new(
+            "constraint-generation",
+            ErrorKind::AssertTypeMismatch("rhs", rhs),
+            span,
```
mimoo, line 81: I think UnexpectedError is still good here, as I don't think we should run into these things (the type checker will catch that before we hit this place)

src/parser/types.rs
```diff
@@ -452,7 +455,7 @@ impl Ty {
                 let ty_kind = Self::reserved_types(module, name);
```
mimoo, line 455: add the `?` here, it's cleaner : o

src/type_checker/checker.rs
```diff
@@ -250,7 +250,11 @@ impl<B: Backend> TypeChecker<B> {
                             .var_name
                             .expect("anonymous lhs access cannot be mutated")
                     }
-                    _ => panic!("bad expression assignment (TODO: replace with error)"),
+                    _ => Err(Error::new(
+                        "compute-type",
+                        ErrorKind::UnexpectedError("bad expression assignment"),
+                        expr.span,
+                    ))?,
```
mimoo, line 257: IIRC the way is to write `Err(self.error(...))` as in other places

mimoo: same for the errors below

src/stdlib/builtins.rs
crypdoughdoteth: The main issue that I ran into here is that I wanted to also include `res_info.typ` in the error message just like in the panic version. format!()` returns `String` and `UnexpectedError` requires `&'static str` as its tagged field. Borrowing the `String` does not last for `'static` unless I leak the allocation. In order to get around this, I created this new error variant. The alternative was a slightly more opaque error message. I'm not sure if this is the right thing per se, but worth discussing briefly

src/type_checker/checker.rs
```diff
@@ -294,28 +294,35 @@ impl<B: Backend> TypeChecker<B> {
...
+                    .get_type_info(&lhs_name)
+                    .or_else(|_| {
+                        Err(self.error(ErrorKind::UnexpectedError("variable not found"), expr.span))
+                    })?
                     .clone();
 
                 // and is mutable
-                if !lhs_info.mutable {
-                    return Err(self.error(ErrorKind::AssignmentToImmutableVariable, expr.span));
+                if let Some(lhs_info) = lhs_info {
```
mimoo, line 312: wait why is lhs_info still an option here

src/type_checker/checker.rs
```diff
@@ -294,28 +294,35 @@ impl<B: Backend> TypeChecker<B> {
...
                 };
 
                 // check that the var exists locally
                 let lhs_info = typed_fn_env
-                    .get_type_info(&lhs_name)?
-                    .expect("variable not found (TODO: replace with error")
+                    .get_type_info(&lhs_name)
+                    .or_else(|_| {
+                        Err(self.error(ErrorKind::UnexpectedError("variable not found"), expr.span))
+                    })?
```
mimoo, line 308: you want `ok_or` here instead

src/type_checker/checker.rs
crypdoughdoteth: I figured out what went wrong here: I didn't handle the `Result` with `?` produced from `get_type_info()`

mimoo:
Lgtm! Thanks so much for the PR!

I'll let @katat merge after looking at https://github.com/zksecurity/noname/pull/209#discussion_r1814037804

katat:
Looks good, thanks!
Btw, any ideas on why the test CI failed? @crypdoughdoteth

crypdoughdoteth:
CI should be green now 🫡

Issue #203, opened 2024-10-14
Server mode for debugging and understanding the compiler


mimoo:
My dream is to be able to run the compiler in server mode, so that we can produce all stages of compilation and json it and visualize that in the browser.

There must be some js library that given some json structure allows us to create a nested clickable thing.

PR #215, opened 2024-10-25
server mode for visual debugging

mimoo:
addresses https://github.com/zksecurity/noname/issues/203

very primitive and clunky interface:

Screenshot 2024-10-25 at 1 54 35 PM

I run it this way in a noname project:

```console
$ cargo run --manifest-path <noname_path>/Cargo.toml -- build --server-mode
```

the changes are pretty straight forward:

- add `--server-mode` option to the build command
- when enabled this runs a server in a thread, which also serves some `index.html` page on `/`
- an optional channel sender is passed to the different compiler passes, which allows the different passes to send their state to the server
- the server also has a channel to allow the compiler to pause and step only when the user wants to resume (that might be unnecessary, but heh)


src/server/mod.rs
```diff
@@ -0,0 +1,208 @@
...
+            }
+        }
+    });
+
+    // build our application with a route
+    let manifest_dir = std::env::var("CARGO_MANIFEST_DIR").unwrap();
+    let assets_dir = std::path::Path::new(&manifest_dir).join("assets");
+
+    let app = Router::new()
+        //        .route("/", get(root))
```
katat, line 121: outdated comment

src/server/mod.rs
```diff
@@ -0,0 +1,208 @@
...
+    let listener = tokio::net::TcpListener::bind("0.0.0.0:3000").await.unwrap();
+    axum::serve(listener, app).await.unwrap();
+}
+
+//
+// Routes
+//
+
+// basic handler that responds with a static string
+async fn root() -> &'static str {
```
katat, line 141: dead code?

src/server/mod.rs
```diff
@@ -0,0 +1,208 @@
...
+            tx: compiler_to_server,
+            rx: compiler_from_server,
+        };
+
+        (handle, server_shim)
+    }
+
+    pub(crate) fn send<T: Serialize>(&self, title: String, state: &T) {
+        let state = serde_json::to_string(&state).unwrap();
+        let state = BASE64_STANDARD.encode(&state);
```
katat, line 62:
```suggestion
        let state = BASE64_STANDARD.encode(state);
```

src/type_checker/mod.rs
```diff
@@ -36,13 +36,33 @@ where
...
 
-#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Hash)]
+#[derive(Debug, Clone, SerializeDisplay, Deserialize, PartialEq, Eq, Hash)]
 pub struct FullyQualified {
     /// Set to `None` if the function is defined in the main module.
     pub module: Option<UserRepo>,
     pub name: String,
 }
 
+// impl ToString for FullyQualified {
```
katat, line 46: remove outdated code?

src/type_checker/mod.rs
```diff
@@ -61,29 +81,56 @@ impl FullyQualified {
     }
 }
 
+pub mod attr {
+    use serde::{Serialize, Serializer};
```
katat, line 85: unused `Serialize`

src/type_checker/mod.rs
```diff
@@ -61,29 +81,56 @@ impl FullyQualified {
...
+
+    // pub type Attr<T: Serialize> = std::collections::HashMap<super::FullyQualified, T>;
```
katat, line 87: remove deadcode?

src/type_checker/mod.rs
```diff
@@ -61,29 +81,56 @@ impl FullyQualified {
     }
 }
 
+pub mod attr {
```
katat, line 84: this whole attr seems unused

src/cli/cmd_build_and_check.rs
```diff
@@ -116,6 +130,9 @@ pub fn cmd_build(args: CmdBuild) -> miette::Result<()> {
 
     println!("successfully built");
 
+    // server mode
+    handle.map(|h| h.join().unwrap());
```
katat, line 134: Since we don't use the return from the map function, maybe be better to this instead:

```suggestion
    if let Some(h) = handle { h.join().unwrap() }
```

src/cli/cmd_build_and_check.rs
```diff
@@ -184,7 +205,7 @@ fn produce_all_asts<B: Backend>(path: &PathBuf) -> miette::Result<(Sources, Type
     let mut tast = TypeChecker::new();
 
     // adding stdlib
-    add_stdlib(&mut sources, &mut tast, node_id)?;
+    add_stdlib(&mut sources, &mut tast, node_id, server_mode)?;
```
katat, line 208: maybe it is better to create server_mode fields in nast/tast/mast structs? so we don't have to add `server_mode` as an argument everytime we call the functions?

src/lexer/tokens.rs
```diff
@@ -25,6 +25,12 @@ impl Tokens {
         }
     }
 
+    /// Don't use this function unless you're testing, it's going to be expensive otherwise.
+    pub fn as_vec(&self) -> Vec<Token> {
+        assert!(self.peeked.is_none());
+        self.inner.clone().into_iter().collect()
```
katat, line 31:
```suggestion
        self.inner.clone().collect()
```

src/parser/mod.rs
```diff
@@ -12,6 +12,8 @@ pub mod structs;
 pub mod types;
 
 pub use expr::{Expr, ExprKind, Op2};
+use serde::Serialize;
+use serde_with::serde_as;
```
katat, line 16: unused?

src/compiler.rs
```diff
@@ -132,27 +159,68 @@ pub fn get_nast<B: Backend>(
...
     tast: TypeChecker<B>,
     backend: B,
+    server_mode: &mut Option<crate::server::ServerShim>,
 ) -> miette::Result<CompiledCircuit<B>> {
+    // monomorphization pass
     let mast = mast::monomorphize(tast)?;
+
+    // debug server
+    if let Some(server_mode) = server_mode {
+        server_mode.send(format!("monomorphization of whole project"), &mast);
```
katat, line 212:
```suggestion
        server_mode.send("monomorphization of whole project".to_string(), &mast);
```

katat (approved):
Very nice feature to have!

src/cli/cmd_build_and_check.rs
mimoo: what's the difference?

src/cli/cmd_build_and_check.rs
mimoo: yeah I think we should have some sort of super `Options` struct so that we can manage all of these arguments that we pass in different places, and have it have a `Default` implementation as well as setters

mimoo:
I'll wait for you @katat to merge the init-stdlib PR first

src/cli/cmd_build_and_check.rs
katat: then clippy won't throw warning :)

katat:
`init-stdlib` is already merged. I think you need to change the base branch.

PR #216, opened 2024-10-28
fix: forbid generic call with mutable vars

katat:
We should also forbid generic function calls with mutable variables that even out of the for loop scope. Otherwise, it will run into unrolling issue.

For example

```rust
fn gen(const LEN: Field) -> [Field; LEN] {
    return [0; LEN];
}

fn loop() {
    let mut size = 2;
    for ii in 0..3 {
        gen(size); // should be forbidden even the var is out of forloop scope
        size = size + 1;
    }
}
```

Even the `size` variable is not within the for loop scope, but it is mutable. 
Similarly, this issue can also happen to struct that is out of for loop scope, but mutable.

### DIFF
--- a/src/type_checker/fn_env.rs
+++ b/src/type_checker/fn_env.rs
@@ -111,10 +111,11 @@ impl TypedFnEnv {
     }
 
     /// Since currently we don't support unrolling, the generic function calls are assumed to target a same instance.
+    /// Each loop iteration should instantiate generic function calls with the same parameters.
     /// This assumption requires a few type checking rules to forbid the cases that needs unrolling.
     /// Forbid rules:
     /// - Access to variables within the for loop scope.
-    /// - Access to mutable variables, except it is an array.
+    /// - Access to mutable variables, except if it is an array.
     ///   Because once the array is declared, the size is fixed even if the array is mutable,
     ///   so the generic value resolved from array size will be same for generic function argument.
     pub fn is_forbidden(&self, scope: usize, ty_info: TypeInfo) -> bool {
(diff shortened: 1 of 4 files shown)


src/type_checker/fn_env.rs
```diff
@@ -110,14 +110,27 @@ impl TypedFnEnv {
         self.current_scope >= prefix_scope
     }
 
-    pub fn is_forbidden(&self, scope: usize) -> bool {
+    /// Since currently we don't support unrolling, the generic function calls are assumed to target a same instance.
```
mimoo, line 113: > each loop iteration should instantiate generic function calls with the same parameters

src/type_checker/fn_env.rs
```diff
@@ -110,14 +110,27 @@ impl TypedFnEnv {
...
+    /// This assumption requires a few type checking rules to forbid the cases that needs unrolling.
+    /// Forbid rules:
+    /// - Access to variables within the for loop scope.
+    /// - Access to mutable variables, except it is an array.
```
mimoo, line 117: why is an array OK?

src/type_checker/fn_env.rs
```diff
@@ -110,14 +110,27 @@ impl TypedFnEnv {
...
+    ///   Because once the array is declared, the size is fixed even if the array is mutable,
```
mimoo, line 118: ah, because generics use the array size, not the array content

mimoo (approved):

src/type_checker/fn_env.rs
katat: yes, exactly. generic use the array size.

katat: note that the type checking will throw error when you try to change the size of the array 
```
arr = truncate(arr); // throws error if the return of truncate() is a different type from arr
```

katat:
Sure, updated the rfc :)

PR #217, opened 2024-10-29
support + - * / % for uints

katat:


src/stdlib/native/int/lib.no
```diff
@@ -97,4 +104,179 @@ fn Uint64.less_than(self, rhs: Uint64) -> Bool {
...
+    return Uint16.new(self.inner - rhs.inner);
+}
+
+fn Uint32.sub(self, rhs: Uint32) -> Uint32 {
+    return Uint32.new(self.inner - rhs.inner);
+}
+
+fn Uint64.sub(self, rhs: Uint64) -> Uint64 {
+    return Uint64.new(self.inner - rhs.inner);
+}
```
mimoo, line 141: so, imagine that we are in a small field that's close to 64-bit, then that subtraction would underflow but still be 64-bit. I'm not sure how we can write code that's considerate of all field sizes... maybe we should add an assert in all of these functions that checks that the field size is >128bit (so that an underflow is always larger than 64 bits)

mimoo (approved):

src/stdlib/native/int/lib.no
katat: agreed. I think we should expose a builtin function to return the number of bits of the field modulus, so the native code can call it.

katat: I added a builtin `check_field_size` and apply it to the uints.

PR #218, opened 2024-10-29
supporting debug var value

katat:
Previously, the log function is to print out the compilation info, such as the type of a variable.

This update is to make it print out the witness value instead, which is more useful for debugging purpose.

For example:
```rust
fn mimc7_hash(values: [Field; LEN], key: Field) -> Field {
    // Initialize with the key.
    let mut res = key;
    // Iterate over each value in the input array.
    for value in values {
        // Update the result with the MIMC hash of the value and the current result.
        res = res + (value + mimc7_cipher(value, res));
        log(res);
    }
    // Return the final accumulated result.
    return res;
}
```

It will print out:
```
span:9904, val: -9648106414739122877149794902860786266420046965962985658208312984077979135149
span:9904, val: 5233261170300319370386085858846328736737478911451874673953613863492170606314
span:9904, val: -4718642457857295476661913076129034983054319651083126928208304929878678658037
```

Currently it only support logging Field type. These info is rough still. 
Later we can make it print out more meta data, such as which file the span belonging to and what type of the var etc.

### DIFF
--- a/src/backends/r1cs/mod.rs
+++ b/src/backends/r1cs/mod.rs
@@ -9,10 +9,10 @@ use itertools::{izip, Itertools as _};
 use num_bigint::BigUint;
 use serde::{Deserialize, Serialize};
 
+use crate::circuit_writer::VarInfo;
 use crate::constants::Span;
 use crate::error::{Error, ErrorKind, Result};
-use crate::helpers::PrettyField;
-use crate::parser::FunctionDef;
+use crate::var::ConstOrCell;
 use crate::{circuit_writer::DebugInfo, var::Value};
 
 use super::{Backend, BackendField, BackendVar};
@@ -224,7 +224,10 @@ where
     /// Constraints in the r1cs.
     constraints: Vec<Constraint<F>>,
     witness_vector: Vec<Value<Self>>,
+    /// Debug information for each constraint.
     debug_info: Vec<DebugInfo>,
+    /// Debug information for var info.
+    log_info: Vec<(String, Span, VarInfo<F, LinearCombination<F>>)>,
     /// Record the public inputs for reordering the witness vector
     public_inputs: Vec<CellVar>,
     /// Record the private inputs for checking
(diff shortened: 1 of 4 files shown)


src/backends/r1cs/mod.rs
```diff
@@ -463,6 +467,21 @@ where
             witness[var.index] = val;
         }
 
+        // print out the log info
+        for (_, span, var_info) in &self.log_info {
```
mimoo, line 471: what about that msg?

src/backends/r1cs/mod.rs
```diff
@@ -463,6 +467,21 @@ where
...
+                    ConstOrCell::Const(cst) => {
+                        println!("span:{}, cst: {}", span.start, cst.pretty());
+                    }
+                    ConstOrCell::Cell(cell) => {
+                        let val = cell.evaluate(&witness);
+                        println!("span:{}, val: {}", span.start, val.pretty());
+                    }
+                }
+            }
+        }
```
mimoo, line 483: I guess that works! So IIUC we just log the vars that we want to print out in order, and then when we generate the witness we go through that.

Maybe it would be worth moving this in the next loop so that we stop printing these if we hit an error in the constraint satisfaction

mimoo (approved):
looks like a great minimal way to make it work! This should allow us to debug a bunch of stuff :)

src/backends/r1cs/mod.rs
katat: currently the native code doesn't support string type, so the msg for each log can't be customized.
I guess this will be only possible after it supported string type or string stdlib.

src/backends/r1cs/mod.rs
katat: > I guess that works! So IIUC we just log the vars that we want to print out in order, and then when we generate the witness we go through that.

Yes :)

This loop is not symmetric with the next loop. So atm it is not straightforward to merge these two loop. Maybe we don't need this once we got the IR.

PR #220, opened 2024-11-04
Support hint function

katat:
This is a POC to support native hint function based on [circ](https://github.com/circify/circ)'s IR and API.

The integration is quite straightforward. The major changes are listed below. Most of the code in `circuit_writer/ir.rs` remain the same as the circuit synthesizer:

1. circ IR can be composed using its `term` API, which can be used independently. (doesn't have to couple with other parts in circ)

2. creates a IR translator based on the template of synthesizer, to translate the mast to IR if it is a hint function.

3. one major change is to replace the `ConstOrCell` in the `Var` with the circ IR term, as a term can represent either constant field or a variable.
eg:

```rust
pub struct Var {
    pub cvars: Vec<Term>,
    pub span: Span,
}
```

So we may want to consider refactor the `var::Var` to cover both `ConstOrCell` and `Term`, in order to avoid the code duplication in the `ir.rs`.

4. another major change is to translate the binary operations or conditional branch.
eg:
```rust
Op2::Division => {
    let t: Term = term![
        Op::PfNaryOp(PfNaryOp::Mul); lhs.cvars[0].clone(), 
        term![Op::PfUnOp(PfUnOp::Recip); rhs.cvars[0].clone()]
    ];
    Var::new_cvar(t, expr.span)
}
```

5. atm, everything is defined as field type at low level using the circ IR

6. create a `Value::HintIR` to represent an IR generated from a hint function, and compute it via `compute_val` using circ's PreComp API. 

TODOs:
- [x] support accepting other noname argument types and return types for hint functions.


src/backends/mod.rs
```diff
@@ -199,6 +201,44 @@ pub trait Backend: Clone {
...
+                let cfg = CircCfg::default();
+                let cfg_f = cfg.field();
+
+                // map the named vars to env
+                let env = named_vars
+                    .iter()
+                    .map(|(name, var)| {
+                        let val = match var {
+                            // todo: convert ark_ff to rug::integer instead of u64
+                            crate::var::ConstOrCell::Const(cst) => cfg_f.new_v(cst.to_u64()),
```
StefanosChaliasos, line 220: Is it safe to convert to u64 here?

src/circuit_writer/ir.rs
```diff
@@ -0,0 +1,1112 @@
...
+
+    pub fn new_var(cvar: Term, span: Span) -> Self {
+        Self {
+            cvars: vec![cvar],
+            span,
+        }
+    }
+
+    pub fn new_constant(cst: BigUint, span: Span) -> Self {
+        let cst: u32 = cst.try_into().expect("expected u32");
```
StefanosChaliasos, line 54: why do we use u32 here?

src/circuit_writer/ir.rs
```diff
@@ -0,0 +1,1112 @@
...
+        let cst_v = cfg_f.new_v(cst);
+        let cvar = leaf_term(Op::new_const(Value::Field(cst_v)));
+
+        Self {
+            cvars: vec![cvar],
+            span,
+        }
+    }
+
+    pub fn new_constant_typ<F: BackendField>(cst_info: &ConstInfo<F>, span: Span) -> Self {
```
StefanosChaliasos, line 68: typo?

src/circuit_writer/mod.rs
```diff
@@ -44,6 +45,8 @@ where
     /// 3. During witness generation, the public output computation
     ///    is delayed until the very end.
     pub(crate) public_output: Option<Var<B::Field, B::Var>>,
+
+    ir_writer: ir::IRWriter<B>,
```
StefanosChaliasos, line 49: This is basically only for hints, right? Should we add a comment?

src/parser/types.rs
```diff
@@ -1133,6 +1133,18 @@ impl FunctionDef {
     pub fn parse_fn_body(ctx: &mut ParserCtx, tokens: &mut Tokens) -> Result<Vec<Stmt>> {
         let mut body = vec![];
 
+        // return empty body when the next token is `;` instead of `{`
```
StefanosChaliasos, line 1136: What is the scenario for that? Do we have test cases?

StefanosChaliasos (commented):
The PR lgtm, I have a few questions wrt the type transformations.

src/backends/mod.rs
katat: This should be field based. This will be fixed.

src/circuit_writer/ir.rs
katat: Save as above. Will be field based.

src/circuit_writer/mod.rs
katat: Yes, will add more documentations.

src/parser/types.rs
katat: Before this PR, the hint functions are implemented in builtins. So the case of empty body will be deprecated.

src/backends/mod.rs
mimoo, line 220: wait why are we truncating the value here?

src/backends/mod.rs
```diff
@@ -199,6 +201,44 @@ pub trait Backend: Clone {
...
+                            crate::var::ConstOrCell::Cell(var) => {
+                                let val = self.compute_var(env, var).unwrap();
+                                // todo: convert ark_ff to rug::integer instead of u64
+                                cfg_f.new_v(val.to_u64())
```
mimoo, line 224: same here

src/circuit_writer/ir.rs
```diff
@@ -0,0 +1,1112 @@
...
+
+                // we already checked for early returns in type checking
+                return Ok(Some(var));
+            }
+        }
+        fn_env.pop();
+        Ok(None)
+    }
+
+    // todo: now only assume it is always field, later it should support other types for both inputs and outputs
```
mimoo, line 532: legacy comment?

src/circuit_writer/ir.rs
katat: it is the coding style :D

mimoo:
LGTM but @StefanosChaliasos I'll let you approve

src/circuit_writer/ir.rs
```diff
@@ -0,0 +1,1119 @@
+use ark_ff::Zero;
```
mimoo, line 1: btw as most of this file will be useful in the future to have an IR pass, I think it should be moved to an circ_ir folder or something, and moved out of the circuit_writer

mimoo (approved):
approving for the sake of getting this moved, let's add a disclaimer that we're still experimental in the README

src/circuit_writer/ir.rs
```diff
@@ -0,0 +1,1119 @@
...
+        fn_env.pop();
+        Ok(None)
+    }
+
+    pub fn compile_hint_function_call(
+        &mut self,
+        function: &FunctionDef,
+        args: Vec<crate::circuit_writer::fn_env::VarInfo<B::Field, B::Var>>,
+    ) -> Result<Vec<crate::var::Value<B>>> {
+        assert!(!function.is_main());
```
StefanosChaliasos, line 532: Do we actually check that we are calling a hint function and not a non hint one?

StefanosChaliasos: Also, a hint should only be able to call hint function, right?

src/circuit_writer/ir.rs
```diff
@@ -660,6 +660,14 @@ impl<B: Backend> IRWriter<B> {
                     FnKind::Native(func) => {
                         // module::fn_name(args)
                         // ^^^^^^
+                        // only allow calling hint functions
+                        if !func.is_hint {
+                            return Err(self.error(
+                                ErrorKind::InvalidFnCall("only hint functions allowed"),
+                                expr.span,
+                            ));
+                        }
```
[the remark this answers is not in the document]
StefanosChaliasos: Ok, solves my previous comment :D. I guess here we check that a hint can call only another hint which makes sense

src/circuit_writer/ir.rs
```diff
@@ -0,0 +1,1119 @@
...
+
+    pub fn is_empty(&self) -> bool {
+        self.cvars.is_empty()
+    }
+
+    pub fn get(&self, idx: usize) -> Option<&Term> {
+        if idx < self.cvars.len() {
+            Some(&self.cvars[idx])
+        } else {
+            None
```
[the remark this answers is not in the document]
katat: yeah, it is currently handled by callers.

src/circuit_writer/ir.rs
```diff
@@ -0,0 +1,1119 @@
...
+                self.add_local_var(fn_env, lhs.value.clone(), var_info)?;
+            }
+
+            StmtKind::ForLoop {
+                var,
+                argument,
+                body,
+            } => {
+                match argument {
+                    ForLoopArgument::Range(range) => {
```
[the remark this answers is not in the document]
katat: yes

src/circuit_writer/ir.rs
```diff
@@ -0,0 +1,1119 @@
...
+                        let start_bg: BigUint = self
+                            .compute_expr(fn_env, &range.start)?
+                            .ok_or_else(|| {
+                                self.error(ErrorKind::CannotComputeExpression, range.start.span)
+                            })?
+                            .constant()
+                            .expect("expected constant")
+                            .into();
+                        let start: u32 = start_bg.try_into().map_err(|_| {
+                            self.error(ErrorKind::InvalidRangeSize, range.start.span)
```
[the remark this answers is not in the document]
katat: atm, the size of the number is restricted to u32, so it throws error if it is bigger.

src/circuit_writer/ir.rs
katat: yes, only hint functions are allowed.

src/circuit_writer/ir.rs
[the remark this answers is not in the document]
StefanosChaliasos: Maybe we should add an issue to change that? Or do you think it's better to handle that in the callers? In any case even if you return a result it will be basically handled by callers but it would be more explicit and it will probably print a nicer error message?

src/circuit_writer/ir.rs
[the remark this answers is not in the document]
StefanosChaliasos: Why is that? I guess when we are in a hint, there isn't any reason why this should be known at compile time as we are out-of-circuit, right? Do I miss something?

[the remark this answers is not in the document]
katat: oh yeah, this is a good point. I created an issue for that: https://github.com/zksecurity/noname/issues/234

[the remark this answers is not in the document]
katat: oh wait, I misunderstood your question.
so yeah, this is an interesting insight. It should allow variables as the range.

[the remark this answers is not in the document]
katat: I guess lets keep this way for now, and will revisit this when someone raises the scenarios they need the range to be variables.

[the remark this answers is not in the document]
mimoo: we don't do TDD here, we do user-complaint driven coding

PR #225, opened 2024-11-09
Add += functionality

### LINKED ISSUE #175: `+=` operator
it would be good to implement a `+=` operator and fix the comment in `examples/iterate.no`

bufferhe4d:
Attempt to resolve #175

### DIFF
--- a/src/type_checker/checker.rs
+++ b/src/type_checker/checker.rs
@@ -389,6 +389,112 @@ impl<B: Backend> TypeChecker<B> {
                     | Op2::Division
                     | Op2::BoolAnd
                     | Op2::BoolOr => lhs_node.typ,
+                    Op2::PlusEqual => {
+                        // todo: better way to do this instead of copying assignment logic?
+                        // compute type of lhs
+                            let lhs_node = self
+                            .compute_type(lhs, typed_fn_env)?
+                            .expect("type-checker bug: lhs access on an empty var");
+
+                        // todo: check and update the const field type for other cases
+                        // lhs can be a local variable or a path to an array
+                        let lhs_name = match &lhs.kind {
+                            // `name = <rhs>`
+                            ExprKind::Variable { module, name } => {
+                                // we first check if it's a constant
+                                // note: the only way to check that atm is to check in the constants hashmap
+                                // this is because we don't differentiate const vars from normal variables
+                                // (perhaps we should)
+                                let qualified = FullyQualified::new(&module, &name.value);
+                                if let Some(_cst_info) = self.const_info(&qualified) {
+                                    return Err(self.error(
+                                        ErrorKind::UnexpectedError("cannot assign to an external variable"),
+                                        lhs.span,
+                                    ));
+                                }
+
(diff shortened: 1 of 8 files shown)


src/type_checker/checker.rs
```diff
@@ -389,6 +389,112 @@ impl<B: Backend> TypeChecker<B> {
                     | Op2::Division
                     | Op2::BoolAnd
                     | Op2::BoolOr => lhs_node.typ,
+                    Op2::PlusEqual => {
+                        // todo: better way to do this instead of copying assignment logic?
```
bufferhe4d, line 393: Here I copied the same logic from ExprKind::Assignment. I am not sure if this is the correct way to do it.

katat: Yeah, this looks redundant to the assignment.

One alternative may be to transform the expression from `lhs += rhs` to `lhs = lhs + rhs`. The transformation would need to be before the type checking phase. Maybe this can happen in the name resolution pipeline: https://github.com/zksecurity/noname/blob/78ff4ba874a67bcf4e6acd7eb831609fbc524683/src/name_resolution/expr.rs#L55

@mimoo wdyt?

katat: @bufferhe4d 
I just had a discussion with @mimoo. We see the `+=` more like a shortcut instead of new operator. We thought it would be better to implement the transformation as described above at the name resolution. That would simplify the PR and avoid the redundancy in the type checking. 

Please feel free to comment if you have any questions.

bufferhe4d: Alright, this looks like a much more elegant approach. Thanks. Few questions:

1) If we decide to do it under name resolution, it feels like either we need to keep a new operator under Op2 to recognize this op and resolve. Or we need a new ExprKind (could be called AugmentedAssignment). I guess you would prefer the latter to get rid of defining a new operation under Op2 Right?
2) If we also wouldn't prefer to create a new ExprKind . Shall I attempt to make this resolution directly in parse_rhs ?
https://github.com/zksecurity/noname/blob/78ff4ba874a67bcf4e6acd7eb831609fbc524683/src/parser/expr.rs#L410

katat: It would be better if we can avoid a new Op2 for this case.

So looks like the `parse_rhs` is the way to go. That is a great insight @bufferhe4d 👍

bufferhe4d: Alright, here is my new attempt. I also went ahead and implemented -= and *= inside `parse_rhs`, added some tests as well. I hope this works as intended now. Please let me know if anything is problematic. Thanks.

katat (approved):
Looks great to me!

katat:
There are CI errors on formatting. You may want to run `cargo fmt` to format the code.

bufferhe4d:
> There are CI errors on formatting. You may want to run `cargo fmt` to format the code.

Indeed, should be good to go now.

katat:
Merged. Thanks again!

mimoo:
woop woop!

PR #231, opened 2024-11-11
Add boolean flag to skip argument type checking of builtin functions.

### LINKED ISSUE #228: make log work with any types
currently the `log` function signature is hardcoded as:

```rust
const LOG_FN: &str = "log(var: Field)";
```

this makes the type checker forbid the function from being used with other types than `Field`.

To fix that, we need to:

1. make it so that the arguments part of a hardcoded builtin signature can be ignored by the type checker
2. while still making sure that the type checker can verify that the returned type of a builtin function matches the expected return type expected by the caller (the builtin function does not have that information so cannot do that type checking)

In `get_parsed_fns` we need a flag that signals this:

```rust
        for (code, fn_handle, ignore_args) in fns {
```

Then `FnKind::BuiltIn` needs to carry that information as well.

Finally, in the typechecker, we can ignore type checks on the function arguments if we see this flag.

bufferhe4d:
Attempts to resolve #228.

Here is my approach:

As suggested in #228 I first introduced a new flag for the `FnKind::Builtin` called `ignore_arg_types`:

https://github.com/bufferhe4d/noname/blob/b243dce51ddbb96babf627fd3c1a8b5266d261ab/src/imports.rs#L78

Propagated this to all the occurrences of this. Then implemented the actual skip in the type checker.

### DIFF
--- a/src/type_checker/checker.rs
+++ b/src/type_checker/checker.rs
@@ -11,12 +11,13 @@ use crate::{
     imports::FnKind,
     parser::{
         types::{
-            is_numeric, ModulePath, FnSig, ForLoopArgument, FunctionDef, Stmt, StmtKind, Symbolic, Ty, TyKind,
+            is_numeric, FnSig, ForLoopArgument, FunctionDef, ModulePath, Stmt, StmtKind, Symbolic,
+            Ty, TyKind,
         },
         CustomType, Expr, ExprKind, Op2,
     },
-    syntax::is_type,
     stdlib::builtins::QUALIFIED_BUILTINS,
+    syntax::is_type,
 };
 
 use super::{FullyQualified, TypeChecker, TypeInfo, TypedFnEnv};
@@ -842,7 +843,7 @@ impl<B: Backend> TypeChecker<B> {
                 &fn_sig.name.value,
             ))
             .map(|info| &info.kind)
-        {   
+        {
             // check builtin
             Some(FnKind::BuiltIn(_, _, ignore)) => *ignore,
             _ => false,
--- a/src/type_checker/checker.rs
+++ b/src/type_checker/checker.rs
@@ -5,16 +5,18 @@ use serde::{Deserialize, Serialize};
 
 use crate::{
     backends::Backend,
+    cli::packages::UserRepo,
     constants::Span,
     error::{ErrorKind, Result},
     imports::FnKind,
     parser::{
         types::{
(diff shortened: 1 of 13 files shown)


src/type_checker/checker.rs
```diff
@@ -833,6 +835,19 @@ impl<B: Backend> TypeChecker<B> {
             None => (),
         };
 
+        // get the ignore_arg_types flag from the function info if it's a builtin
+        let ignore_arg_types = match self
```
bufferhe4d, line 839: Seems to work for me but I am not sure if this is the correct way to check if a function is builtin.

src/type_checker/mod.rs
```diff
@@ -392,7 +392,8 @@ impl<B: Backend> TypeChecker<B> {
                             qualified,
                             FnInfo {
                                 is_hint: true,
-                                kind: FnKind::BuiltIn(function.sig.clone(), fn_handle),
+                                // todo: is there a case where we want to ignore argument types for hint functions?
+                                kind: FnKind::BuiltIn(function.sig.clone(), fn_handle, false),
```
bufferhe4d, line 396: I couldn't immediately decide what to do here since we first convert a hint function to a builtin one. It felt like we would never disable argument type checking for a hint function so I defaulted to false.

katat: The builtin hints will be deprecated soon. So don't worry bout this. 
Good insight!

katat (approved):
lgtm!
@mimoo may want another pass

mimoo:
maybe we should check if this allows us to implement what we wanted to implement on top of that?

@varunthakore do you want to try to implement a generic `log()` function on top of this branch to see if it works?

@bufferhe4d if you want to do something similar, a generic `assert_eq()` function would be great, as long as it checks that the two types passed as arguments match

bufferhe4d:
> @bufferhe4d if you want to do something similar, a generic `assert_eq()` function would be great, as long as it checks that the two types passed as arguments match

I currently have an assert_eq that is working with all the types (Field, Bool, a7rbitrary depth structs and arrays) in the language (using the boolean flag we introduced in this PR). It only checks whether the two argument types matched as you described.

Since it requires the flag included here, should I push this functionality to this branch to make it a part of this PR or should I wait until this is merged to open a separate PR?

mimoo:
can you open a separate PR that targets this branch? Or another PR that we can easily rebase?

bufferhe4d:
I created a draft PR from my generic_assert_eq here: https://github.com/bufferhe4d/noname/pull/2

The CI should be fixed now.

Issue #134, opened 2024-06-20
be clearer on mutability


mimoo:
we're not super clear on what can be mutated and what cannot be.

For example, can this function call mutate `thing`?

```rust
let mut thing = 5;
stuff(thing);
```

imo we should make mutability when we're passing mutable values around. In rust we would have to write `stuff(&mut thing)`. In noname everything is passed by value so I'm not sure what would be a good syntax (there is no such thing as `stuff(&thing)` for example. Maybe `stuff(mut thing)`?

How about this example:

```rust
fn stuff(arg: Field) { /* ... */ }
```

can the body of `stuff` mutate `arg`? I think it could only if we have `fn stuff(mut arg: Field)`, which I don't think is something we currently have in functions/methods.

The best thing to do would be to create tests to showcase these examples

katat:
`fn stuff(mut arg: Field)` makes sense to mutate an external variable inside the function. 

atm, all the variables passing between the functions are done via copying their values. The reference variable is only usable at local scope.

mimoo:
arg, so now I'm wondering what this does exactly if `stuff` mutates `a`:

```rust
let mut a = 5;
stuff(a);
```

katat:
The function `stuff(a)` can be called with that mutable variable. Inside that function, that variable `a` is seen as immutable. As long as it doesn't directly mutate the argument `a` which is not mutable as the syntax `mut` is not supported for the function argument, it won't throw error.

bufferhe4d:
I faced this problem when I was planning to test the example given in #224 . 

Below, even if the room variable is mut, we are not able to write functions like update_size since it is not possible to pass self as mut. I don't know if this is necessary to have but just found it relevant to share.

```
struct Room {
	pub beds: Field, // public
        size: Field // private
}

Room.update_size(self, size: Field) {
        self.size = size; 
        ^ error: self is not mutable
}

fn main() {
       let mut room = Room {beds: 2, size: 10};
       room.beds = 2; // allowed
       room.size = 5;  // not allowed
       room.update_size(5); // allowed
}
```

mimoo:
I think we're in a model where functions/methods have to produce a new object instead of updating the current object. It's more "functional" and doesn't really save you any constraints underneath so maybe we should keep it this way?

mimoo:
[...]

PR #232, opened 2024-11-12
Add pub attribute for struct fields

### LINKED ISSUE #224: Allow pub attribute for struct fields
For struct fields, sometimes we want to control the accessibility of the fields, so they are only accessible within the struct methods.

For example: 

```rust
struct Room {
	pub beds: Field, // public
        size: Field // private
}

Uint8.update_size(self, size: Field) {
        self.size = size; 
}

fn main() {
       let room = Room {beds: 2, size: 10};
       uint.beds = 2; // allowed
       uint.size = 5;  // not allowed
       uint.update_size(5); // allowed
}
```

When there is a `pub` attribute for a struct field, the field can be accessible from external. Otherwise, the private fields are only accessible in the struct methods.

bufferhe4d:
Attempts to resolve #224 

Here is my approach:

I introduced a new component  `Option<Attribute>` to `fields` in order to make use of the already defined `Pub` attribute that is used for public inputs.

The tricky part was to check whether we are inside a structs method or not. For that I introduced a new variable inside  `TypedFnEnv` called `current_fn_kind `  that keeps track of the function we are in. This is further utilized in the type checker.


katat (approved):
Nice!
It would be great to support modifying the struct fields inside the methods, as this `pub` is also about a whitelist to controlling the values of a struct.

Maybe we can create another PR to implement the `mut [arg]` as described [here](https://github.com/zksecurity/noname/issues/134).

src/type_checker/checker.rs
```diff
@@ -116,14 +117,39 @@ impl<B: Backend> TypeChecker<B> {
...
                     .iter()
-                    .find(|(name, _)| name == &rhs.value)
-                    .map(|(_, typ)| typ.clone());
+                    .find(|(field_name, _, _)| field_name == &rhs.value)
+                {
+                    // check for the pub attribute
+                    let is_public = attribute
+                        .as_ref()
+                        .map(|attr| matches!(attr.kind, AttributeKind::Pub))
+                        .unwrap_or(false);
```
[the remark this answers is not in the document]
mimoo: (btw to avoid the option we could also have a `AttributeKind::None`)

src/type_checker/checker.rs
```diff
@@ -116,14 +117,39 @@ impl<B: Backend> TypeChecker<B> {
...
+                    let in_method = if let Some(fn_kind) = typed_fn_env.current_fn_kind() {
+                        match fn_kind {
+                            FuncOrMethod::Method(method_struct) => {
+                                method_struct.name == struct_name
+                            }
+                            FuncOrMethod::Function(_) => false,
+                        }
+                    } else {
+                        false
+                    };
```
mimoo, line 141: there's gotta be a way to decrease the LOC here as well. Maybe

```rust
let in_method = matches!(typed_fn_env.fn_kind, FuncOrMethod::Method(struc) if struc.name == struct_name);
```

mimoo (commented):
left a few more comments! Sorry for the delay

src/negative_tests.rs
```diff
@@ -747,3 +747,30 @@ fn test_mut_cst_struct_field_prop() {
...
+    }
+
+    fn Room.access_size(self) {
+        let room2 = Room {beds: self.beds, size: self.size};
+    }
+
+    fn main(pub beds: Field) {
+        let mut room = Room {beds: 2, size: 10};
+        room.beds = beds; // allowed
+        room.size = 5; // not allowed
```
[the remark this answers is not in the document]
mimoo: agree ^ we can simplify the test

src/parser/structs.rs
```diff
@@ -55,6 +55,26 @@ impl StructDef {
                 tokens.bump(ctx);
                 break;
             }
+
```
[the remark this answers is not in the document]
mimoo: at the same time it seems OK to allow it...

src/type_checker/fn_env.rs
```diff
@@ -55,6 +55,9 @@ pub struct TypedFnEnv {
 
     /// Determines if forloop variables are allowed to be accessed.
     forbid_forloop_scope: bool,
+
+    /// The kind of function we're currently type checking
+    current_fn_kind: Option<FuncOrMethod>,
```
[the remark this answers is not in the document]
bufferhe4d: For some reason, I thought we have global scope in the language and we might not be in a function while type checking. Should be fixed now.

bufferhe4d:
Alright, I did another iteration after the comments from @mimoo. I have one question about how we initialize the TypedFnEnv now below.

src/type_checker/fn_env.rs
```diff
@@ -57,13 +57,19 @@ pub struct TypedFnEnv {
...
-    current_fn_kind: Option<FuncOrMethod>,
+    current_fn_kind: FuncOrMethod,
 }
 
 impl TypedFnEnv {
-    /// Creates a new TypeEnv
-    pub fn new() -> Self {
-        Self::default()
+    /// Creates a new TypeEnv with the given function kind
+    pub fn new(fn_kind: &FuncOrMethod) -> Self {
```
bufferhe4d, line 65: I am sure there is a more elegant way of doing this. The reason I went with this is that since current_fn_kind is not an option, calling Self::default() calls the default constructor of FuncOrMethod as well, which calls unreachable!:

https://github.com/zksecurity/noname/blob/b17ea203e24552974c499ee17bae3fa85b044247/src/parser/types.rs#L739

How can we do it better?

katat: Good catch.
With `FuncOrMethod`, I think it should be always calling ::new to specify the `fn_kind`.

bufferhe4d: I think I didn't fully understand the last part. Do you mean the default() for `FuncOrMethod` should change (return a new object instead of unreachable) or is the current approach okay as is?

katat: I meant default() should be removed, if it is not used anymore.

katat:
It looks good. Once the minor conflict is fixed, I think we can merge it :)

mimoo:
can we merge this one :D? cc @bufferhe4d sorry for the delay!

katat:
Closing this one, as I create a PR https://github.com/zksecurity/noname/pull/263 to fix the pending issues on this one. Thanks again @bufferhe4d !

PR #235, opened 2024-11-14
Add --disable-safety-check option

StefanosChaliasos:
This option allows variables that are not constrained.

Right now, it's a bit ugly how this parameter has to be passed. I will explain the workflow below, but I guess in the future, we can have a struct to capture parameters and pass them around so we can have access to user-defined configs. 

We need to add some ifs for this particular feature when we validate that variables have been used. This is happening in the `finalize_circuit` function inside circuit_writer instead of in the type checking and semantic checks. To pass the parameter from the UI based on which we are going to return an error or not, we do the following:

1. Add the option to `cli/cmd_build_and_check.rs` and `cli/cmd_prove_and_verify.rs`
2. Add bool parameter (`disable_safety_check`) in `compiler.rs::compile` function
3. Pass that parameter to `CircuitWriter::generate_circuit` function
4. Pass that parameter to `finalize_circuit` function (In all backends)
5. Add the if statements (In all backends)

Furthermore, in the tests, we always set that option to false.

### DIFF
--- a/src/backends/kimchi/mod.rs
+++ b/src/backends/kimchi/mod.rs
@@ -358,25 +358,21 @@ impl Backend for KimchiVesta {
         }
 
         for var in 0..self.next_variable {
-            if !written_vars.contains(&var) {
+            if !written_vars.contains(&var) && !disable_safety_check {
                 if let Some(private_cell_var) = self
                     .private_input_cell_vars
                     .iter()
                     .find(|private_cell_var| private_cell_var.index == var)
                 {
                     // TODO: is this error useful?
-                    if !disable_safety_check {
-                        let err = Error::new(
-                            "constraint-finalization",
-                            ErrorKind::PrivateInputNotUsed,
-                            private_cell_var.span,
-                        );
-                        Err(err)?;
-                    }
+                    let err = Error::new(
+                        "constraint-finalization",
+                        ErrorKind::PrivateInputNotUsed,
+                        private_cell_var.span,
+                    );
+                    Err(err)?;
                 } else {
-                    if !disable_safety_check {
-                        Err(Error::new("contraint-finalization", ErrorKind::UnexpectedError("there's a bug in the circuit_writer, some cellvar does not end up being a cellvar in the circuit!"), Span::default()))?;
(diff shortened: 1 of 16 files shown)


src/backends/r1cs/mod.rs
```diff
@@ -446,17 +447,21 @@ where
                     .iter()
                     .find(|private_cell_var| private_cell_var.index == index)
                 {
-                    Err(Error::new(
-                        "constraint-finalization",
-                        ErrorKind::PrivateInputNotUsed,
-                        private_cell_var.span,
-                    ))?
+                    if !disable_safety_check {
```
katat, line 450: same as above

src/backends/kimchi/mod.rs
```diff
@@ -364,14 +365,18 @@ impl Backend for KimchiVesta {
                     .find(|private_cell_var| private_cell_var.index == var)
                 {
                     // TODO: is this error useful?
-                    let err = Error::new(
-                        "constraint-finalization",
-                        ErrorKind::PrivateInputNotUsed,
-                        private_cell_var.span,
-                    );
-                    Err(err)?;
+                    if !disable_safety_check {
```
katat, line 368: Maybe checking this flag at the root block is cleaner :)

```
    if !written_vars.contains(&var) &&  !disable_safety_check {
```

katat (commented):
Cool, lgtm. @mimoo may want to have a pass.

src/cli/cmd_build_and_check.rs
```diff
@@ -71,6 +71,10 @@ pub struct CmdBuild {
     /// Run in server mode to help debug and understand the compiler passes.
     #[clap(long)]
     server_mode: bool,
+
+    /// Do not check that every variable is in a constraint
+    #[arg(long = "disable-safety-check", global = true)]
```
katat, line 76: Maybe adding a doc on why this is necessary in certain cases.

StefanosChaliasos:
@katat, I think it should be good to go now

katat (approved):
Cool, thanks!

PR #238, opened 2024-11-25
Implement generic log

varunthakore:
This PR implements logging of different types for R1CS backend. It is based on PR #231. Currently, the `log` function supports only single argument.


mimoo:
This is great! Can you screenshot an example to see that it's working? Also what is the error when you pass multiple arguments? I thought it would work :o

varunthakore:
> This is great! Can you screenshot an example to see that it's working? Also what is the error when you pass multiple arguments? I thought it would work :o

These screenshots are generated by running tests from `src/tests/example.rs`.

Log Field and Bool.
Screenshot 2024-11-25 at 5 30 49 PM

Log Array.
Screenshot 2024-11-25 at 5 29 27 PM

Log Custom types.
Screenshot 2024-11-25 at 5 32 36 PM

When you pass multiple arguments to `log()` it throws error `MismatchFunctionArguments` because of the argument length check at the following places:
1. https://github.com/varunthakore/noname/blob/c6f240a5ed803073f2d5c8809412cad9aa7a4fb6/src/mast/mod.rs#L1321-L1327
2. https://github.com/varunthakore/noname/blob/c6f240a5ed803073f2d5c8809412cad9aa7a4fb6/src/type_checker/checker.rs#L873-L879

I think these checks cannot be skipped because it will be required for the generic `assert_eq()` function.

src/backends/r1cs/mod.rs
```diff
@@ -478,15 +479,80 @@ where
...
+                        }
+                        ConstOrCell::Cell(cell) => {
+                            let val = cell.evaluate(&witness) == F::one();
+                            println!("{dbg_msg}{}", val);
+                        }
+                    }
+                }
+
+                // Array
+                Some(TyKind::Array(_, _)) => {
```
katat, line 508: Also it could be recursive array.

For example 
1. [[1, 2],[3, 4]]
2. [thing1, thing2] // thing* are structs

So maybe it would be better to recursively print out the elements based on the element's type.

katat (commented):
This looks great!
I left a couple comments. It would be great to recursively print out the array or struct.

src/backends/r1cs/mod.rs
varunthakore: Log Array Array Field.
array_array_field


Log Array Array Bool.
array_array_bool


Log Array Custom.
array_custom

katat (commented):
Looks great!
I add a couple minor comments. Also please resolve the code conflicts so it can be merged.

katat (approved):
Cool, thanks!

PR #240, opened 2024-12-03
Test monomorphized AST

### LINKED ISSUE #200: Test monomorphized AST
After monomorphization phase, it will generate a monomorphized AST (MAST). 
To ensure the monomorphization process works as expected, it'd better to have snapshot tests for the generated MASTs for different cases.

cyberbono3:
Closes #200

### DIFF
--- a/src/backends/r1cs/mod.rs
+++ b/src/backends/r1cs/mod.rs
@@ -258,7 +258,7 @@ where
 }
 
 /// R1CS backend with bls12_381 field.
-#[derive(Clone)]
+#[derive(Clone, Debug)]
 pub struct R1CS<F>
 where
     F: BackendField,
--- a/src/error.rs
+++ b/src/error.rs
@@ -342,6 +342,9 @@ pub enum ErrorKind {
     #[error("cannot compile a module without a main function")]
     NoMainFunction,
 
+    #[error("main function not found")]
+    MainFunctionNotFound,
+
     #[error("invalid hexadecimal literal `${0}`")]
     InvalidHexLiteral(String),
 
--- a/src/mast/mod.rs
+++ b/src/mast/mod.rs
@@ -467,11 +467,15 @@ pub fn monomorphize<B: Backend>(tast: TypeChecker<B>) -> Result<Mast<B>> {
     let mut ctx = MastCtx::new(tast);
 
     let qualified = FullyQualified::local("main".to_string());
-    let mut main_fn = ctx
-        .tast
-        .fn_info(&qualified)
-        .expect("main function not found")
-        .clone();
+
+    let mut main_fn = match ctx.tast.fn_info(&qualified) {
+        Some(fn_info) => fn_info.clone(),
+        None => Err(Error::new(
+            "Backend - Monomorphize",
+            ErrorKind::MainFunctionNotFound,
+            Span::default(),
+        ))?,
+    };
 
     let mut func_def = match &main_fn.kind {
         // `fn main() { ... }`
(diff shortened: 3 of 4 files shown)


katat:
hmm, I think the issue is to add tests to compare the AST during monomorphization phase (MAST). It is useful to review the effects on the generated MAST nodes if there are changes to the logic in MAST pipeline.

There is some tool like https://github.com/mitsuhiko/insta?tab=readme-ov-file to do the job. But I think we should be able to structure the snapshot tests small enough, so the generated AST string can be compared in a manageable way.

cyberbono3:
@katat  Thanks for clarification.  `test_simple_mast` yields some ambiguity in  MAST snapshots. Every new snapshot generates new functions description: `"functions":[["std/bits/nth_bit",` `"functions":[["std/builtins/log"`
You can reproduce it by `cargo insta test --review`

katat:
Ah, it is because the resulted MAST can have the same set of the functions but in different order. The whole MAST contains all the stdlib and builtin functions.

For this kind of test, we should only snapshot the functions in concern, such as the generic examples `test_generic_*`. The snapshot test for MAST is to check if the monomorphization done properly for the generic features.

I think you can create a new test set, similar to the `negative_tests.rs`, for snapshot tests for MAST generation. For the generic feature examples to test snapshot, you can copy the code from the examples folder.

cyberbono3:
> For this kind of test, we should only snapshot the functions in concern, such as the generic examples `test_generic_*`. The snapshot test for MAST is to check if the monomorphization done properly for the generic features.

There are `test_generic_` test cases presented n `src/negative_tests.rs` and in `src/tests/examples.rs`.
I assume, you would like to make snapshot tests for MAST generation  for test cases `test_generic_` listed in  `src/tests/examples.rs` Right?

katat:
> There are test_generic_ test cases presented n src/negative_tests.rs and in src/tests/examples.rs.
I assume, you would like to make snapshot tests for MAST generation for test cases test_generic_ listed in src/tests/examples.rs Right?

Yes, it should only cover the generic functions that involve MAST generation.

> The snapshot test cases from examples folder (arithmetic.no, array.no etc) will produce an ambiguity as I mentioned above.

Not sure If I understood this. The discrepancy is due to the order of the MAST functions. That doesn't mean there is a discrepancy from the MAST. When it comes to comparison, I would suggest checking the monomorphized functions / instantiated function in a narrowed context, that would be easier to find the root causes if there are any discrepancies.

For example, it would be easier to isolate the snapshot test per generic feature, such as generic array:
```
fn clone(const LEN: Field, val: Field) -> [Field; LEN] {
    let mut arr = [0; LEN];
    for idx in 0..LEN {
        arr[idx] = val;
    }
    return arr;
}
```

In other words, in the snapshot test, it should check if the MAST of instantiated function from this generic function is as expected.

src/tests/snapshots/mast.rs
```diff
@@ -0,0 +1,305 @@
...
+        let array = [[xx + 1, xx + 2, xx + 3]; 3];
+        let elm = last(array);
+        assert_eq(elm, 4);
+    }
+    "#;
+    test_monomorphization(code);
+}
+
+#[test]
+fn test_generic_builtin_bits() {
```
cyberbono3, line 66: This tests fails [here](https://github.com/cyberbono3/noname/blob/test-mast/src/mast/mod.rs#L596)

cyberbono3: Basically, I have performed snapshot testing for all fields of `Mast(pub TypeChecker). `

cyberbono3:
Basically, I have performed snapshot testing for all fields of `Mast(pub TypeChecker). ` for generic code snippets.

cyberbono3:
@katat  could you take a look , please?

PR #241, opened 2024-12-08
support left shift or right shift operator

### LINKED ISSUE #181: Support << left shift or >> right shift operator
It is common to need to `<<` or `>>` to manipulate a value via bit shifting.

Without bit shifting, it needs to do the following:
```rust
   // 1 << LEN
    let mut pow2 = 1;
    for ii in 0..LEN {
        pow2 = pow2 + pow2;
    }
```

The left hand side will be always the value, while the right hand side will be the number of bits to shift.

We can add a builtin to support this operator. For example, the compiler will call the newly added builtins corresponding to the following new binary operations. 

```diff
pub enum Op2 {
    Addition,
    Subtraction,
    Multiplication,
    Division,
    Equality,
    Inequality,
    BoolAnd,
    BoolOr,
+   RightShift,
+   LeftShift,
}
```

To implement, it needs to take care two cases for the value being shifted: 
- constant: When the value is a constant, then the builtin function just do the shifting operation and return the value.
- cell var: When the value is a cell var, the builtin needs to handle constraints.

Lynette7:
Resolves #181 

To implement right-shift and left-shift operators, here are the changes I've made:

- Added token recognition for `<<` and `>>` in the lexer
- Extended `Op2` enum to include `RightShift` and `LeftShift`
- Updated the parser to recognize and handle bit shift tokens
- Modified type checker to support the shift operations
- Implemented circuit writer logic for bit shifts using field arithmetic
- Implemented bit shift operations for field elements, supporting constant value shifts and cell variable shifts with constant shift amounts
- Added bitwise shift operations using logical shift right (Lshr) and shift left (Shl) in circuit IR generation

### DIFF
--- a/src/circuit_writer/ir.rs
+++ b/src/circuit_writer/ir.rs
@@ -1,6 +1,6 @@
 use ark_ff::Zero;
 use circ::{
-    ir::term::{leaf_term, term, BoolNaryOp, Op, PfNaryOp, PfUnOp, Sort, Term, Value},
+    ir::term::{leaf_term, term, BoolNaryOp, BvBinOp, Op, PfNaryOp, PfUnOp, Sort, Term, Value},
     term,
 };
 use num_bigint::BigUint;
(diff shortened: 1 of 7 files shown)


katat:
Thanks for addressing this.

I realized this feature can be complicated to support in the constraint level, because the unit in that level is finite field. It makes sense to support logical shifting in finite field, but that can be confusing from user perspective since they might assume they are arithmetic shifting.

Therefore, I think for now we can just make it support in the hint function. That is the feature can be implemented in the `ir.rs`, which is based on Circ term.

Also please add tests in the examples.

katat:
This feature for hint functions has been implemented in another PR https://github.com/zksecurity/noname/pull/253
Thanks a lot for attempting this!

Issue #138, opened 2024-06-24
update to arkworks 0.4.0


mimoo:
https://github.com/zksecurity/noname/pull/137 had to downgrade to 0.3.0, let's try to move to arkworks 0.4.0!

Autoparallel:
I could take a stab at this!

Autoparallel:
Curious what your thoughts are:

The naive approach of "just increment the version" yields some errors that point to the fact that `kimchi` may need a bump as well, unless we want to create some wrappers (which could work, but seems a bit iffy).

Namely: https://github.com/zksecurity/noname/blob/cf9af0b07c248561985fa13107d267655e3619de/src/backends/kimchi/mod.rs#L35-L42

We'll get:
```
the trait bound `ark_ff::fields::models::Fp256<kimchi::mina_curves::pasta::fields::FpParameters>: ark_ff::PrimeField` is not satisfied
```

Stepping into `kimchi`, it's using `ark_ff::PrimeField` from `0.3.0` and implements `PrimeField` on `Fp256` via the macro: 
```rust
impl_Fp!(
    Fp256,
    Fp256Parameters,
    BigInteger256,
    BigInteger256,
    4,
    "256"
);
```

Is it worth trying to push up to `0.4.0` inside of `kimchi` prior to making the change here? Do we instead want to do:
```rust
pub struct VestaField(kimchi::mina_curves::pasta::Fp);
```
and implement the necessary traits on this wrapper? This solution seems less useful as it will likely just be undone if `kimchi` bumps arkworks.

katat:
Nice analysis @Autoparallel !

I think we don't know when kimchi will bump the arkworks version. The wrapper approach you described seems to be way to go. If this wrapper works well, it would become the "template" solution to similar issues in the future as we may add new backend dependencies which we don't have control of.

How many traits does it have to implement? (We may need to evaluate if it is too much work to do so also.)

Autoparallel:
@katat, I don't have the number of traits to impl off the top of my head, but I don't think it is too many. Also, they will likely be pretty easy to implement.  My worry would be a bump in `kimchi` leads to undoing all of this.

If possible, poking at `kimchi` itself to push for a bump may alleviate more pain for everyone down the road. I'm happy to go there and chat about it too!

mimoo:
there was a PR for that but it was never merged: https://github.com/o1-labs/proof-systems/pull/998

Autoparallel:
Oof, and it is quite old. 

[...]

PR #246, opened 2024-12-13
Upgrade to arkworks 0.4.0/2

lopeetall:
Kimchi was upgraded to a newer version that uses Arkworks 0.4, and Arkworks dependencies were upgraded to match those used by Kimchi (either 0.4.0 or 0.4.2 depending).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ camino = "1.1.1" # to replace Path and PathBuf
 clap = { version = "4.0.5", features = ["derive"] } # CLI library
 dirs = "4.0.0" # helper functions (e.g. getting the home directory)
 itertools = "0.10.3" # useful iter traits
-kimchi = { git = "https://github.com/o1-labs/proof-systems", rev = "483bd5bd918ebbc1d5e008ce892d7a0023b50ed1" } # ZKP
+kimchi = { git = "https://github.com/o1-labs/proof-systems"} # ZKP
 miette = { version = "5.0.0", features = ["fancy"] }                                 # nice errors
 num-traits = "0.2.15"                                                                # useful traits on big ints
 once_cell = "1.15.0"                                                                 # for lazy statics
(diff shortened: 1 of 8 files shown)


lopeetall:
Addresses #138

Cargo.toml
```diff
@@ -32,8 +32,7 @@ camino = "1.1.1" # to replace Path and PathBuf
 clap = { version = "4.0.5", features = ["derive"] } # CLI library
 dirs = "4.0.0" # helper functions (e.g. getting the home directory)
 itertools = "0.10.3" # useful iter traits
-kimchi = { git = "https://github.com/o1-labs/proof-systems", rev = "a5d8883ddf649c22f38aaac122d368ecb9fa2230" } # ZKP - Dec 5th, 2023 revision
-#kimchi = { git = "https://github.com/o1-labs/proof-systems", rev = "b9589626f834f9dbf9d587e73fd8176171231e90" } # ZKP
+kimchi = { git = "https://github.com/o1-labs/proof-systems", rev = "483bd5bd918ebbc1d5e008ce892d7a0023b50ed1" } # ZKP
```
mimoo, line 35: any reason for this 3-month old commit?

mimoo:
minimal changes :D nice, (cc @dmpierre I think you were stuck on this one)

Cargo.toml
lopeetall: No reason other than it was the commit referenced in #138. I checked this morning and upgrading to the latest version of kimchi only required a couple more changes, so I went ahead and made them

Cargo.toml
```diff
@@ -32,7 +32,7 @@ camino = "1.1.1" # to replace Path and PathBuf
 clap = { version = "4.0.5", features = ["derive"] } # CLI library
 dirs = "4.0.0" # helper functions (e.g. getting the home directory)
 itertools = "0.10.3" # useful iter traits
-kimchi = { git = "https://github.com/o1-labs/proof-systems", rev = "483bd5bd918ebbc1d5e008ce892d7a0023b50ed1" } # ZKP
+kimchi = { git = "https://github.com/o1-labs/proof-systems"} # ZKP
```
mimoo, line 35: btw this won't build nicely if we don't specify a version here, as it will always pull the latest which might stop working at some point, maybe point to `5b4ac1437e7912237be88d97b4b4891b22e3e61f`?

mimoo (approved):
woop woop, thanks so much for this!

PR #247, opened 2024-12-16
Implement generic assert_eq

bufferhe4d:
Adds generic assert_eq using #231. 

It recursively goes through any nested structure like arrays or structs.


katat (approved):
This is really nice to have! Great work @bufferhe4d !

katat:
@mimoo might want a round of review

mimoo:
let's merge this one?

src/stdlib/builtins.rs
```diff
@@ -33,13 +34,118 @@ impl Module for BuiltinsLib {
...
+            match (lhs_var, rhs_var) {
+                (ConstOrCell::Const(a), ConstOrCell::Const(b)) => {
+                    comparisons.push(Comparison::Constants(a.clone(), b.clone()));
+                }
+                (ConstOrCell::Const(cst), ConstOrCell::Cell(cvar))
+                | (ConstOrCell::Cell(cvar), ConstOrCell::Const(cst)) => {
+                    comparisons.push(Comparison::VarConst(cvar.clone(), cst.clone()));
+                }
+                (ConstOrCell::Cell(lhs), ConstOrCell::Cell(rhs)) => {
+                    comparisons.push(Comparison::Vars(lhs.clone(), rhs.clone()));
```
mimoo, line 78: why return a comparisons array btw? Why not just do the constraints right here?

```
match (lhs_cvar, rhs_cvar) {
        // two constants
        (ConstOrCell::Const(a), ConstOrCell::Const(b)) => {
            if a != b {
                Err(Error::new(
                    "constraint-generation",
                    ErrorKind::AssertionFailed,
                    span,
                ))?
            }
        }

        // a const and a var
        (ConstOrCell::Const(cst), ConstOrCell::Cell(cvar))
        | (ConstOrCell::Cell(cvar), ConstOrCell::Const(cst)) => {
            compiler.backend.assert_eq_const(cvar, *cst, span)
        }
        (ConstOrCell::Cell(lhs), ConstOrCell::Cell(rhs)) => {
            compiler.backend.assert_eq_var(lhs, rhs, span)
        }
```

bufferhe4d: Sorry for the delayed reply. I think the reason was that I couldn't figure how to pass the compiler backend recursively. Hence used the table instead. I can look into it again to make it more elegant.

Issue #248, opened 2024-12-30
Support statements in ITE


katat:
In order to support the following calculation:
```
// 1 if true, 0 if false
function long_gt(n, k, a, b) {
    for (var i = k - 1; i >= 0; i--) {
        if (a[i] > b[i]) {
            return 1;
        }
        if (a[i] < b[i]) {
            return 0;
        }
    }
    return 0;
}
```

The compiler should support statements and early returns in the ITE block, otherwise it is too tricky to implement the logic like this. Perhaps at least we should enhance this feature in the hint functions.

mimoo:
agree that in the hint functions we should allow anything (early return, early break, large if-then-else statements, etc.)

in the rest IMO we should not for now, as it would be hard to support other backends than r1cs

0xnullifier:
@katat I see the pr that you raised was closed for some reason. Is this something not desired anymore?  imho it will be a great feature or is it just too difficult to implement atm?

katat:
It can be tricky to support this on different arithmetic backends. Also there will be issues when calling builtin functions like `assert()`, which is not compatible with this kind of statement support.

However, it should be feasible to enhance the ITE in hint functions, which doesn't involve constraints.

0xnullifier:
Got it so we can do this for hint functions atm. Can I take up this with #211 as they are kind of similar and make a single PR for both or maybe first do #211 and then come back to this?

katat:
Yeah, cool! Let's ignore #211 for now, just focus on enhancing the ITE in hint functions.

0xnullifier:
@katat a question  
For only allowing if else in `hint` functions I would have to know when parsing the statement that whether it was in an hint function or not. So for example this is the sig of `Stmt::parse`
```rust
 pub fn parse(ctx: &mut ParserCtx, tokens: &mut Tokens) -> Result<Self> 
```
so do we change this to
```rust
pub fn parse(ctx: &mut ParserCtx, tokens: &mut Tokens, from_hint: bool) -> Result<Self>
```
or is there any way for me to get the context of body is from which function that I am parsing? I cannot reverse iterate over tokens as the iterator is already moved forward. 
So do we add a field in `ParserCtx` indicating that if it was a hint function or not? or otherwise add a boolean field to `FunctionDef::parse_fn_body` , `Stmt::parse`

0xnullifier:
[...]

PR #249, opened 2025-01-01
Adds sparse merkle tree in noname `stdlib`

### LINKED ISSUE #223: [stdlib] add sparse merkle tree in
We can implement the version using the mimc hash stdlib. 
For reference, see the circom implementation: https://github.com/iden3/circomlib/blob/a8cdb6cd1ad652cca1a409da053ec98f19de6c9d/circuits/smt/smtprocessor.circom#L141

utkarshdagoat:
This pr adds the `sparse merkle tree` to `stdlib` and closes #223 

The testing was done against the output of `circomlibjs` you can find the script [here](https://github.com/utkarshdagoat/smt-test-script). 

The approach taken is almost identical to the circom's implementation mentioned in the issue.


utkarshdagoat:
Hey @katat can you take a look?

katat:
Thanks! Will take a look soon.

utkarshdagoat:
hey  it seems the checks failed due to enforce formating
So I will go ahead and fix them

src/stdlib/native/smt/lib.no
```diff
@@ -0,0 +1,446 @@
...
+/// calculates the next state
+/// # Parameters
+///     `prev_level_state`: state of the previous level
+///     `xor`:  the xor value for the old key  and new key if same we just add a sibling
+///     `is_old_zero`: indicates whethere we are inserting in an zero leaf
+///     `level_inserted`: indicates that is this the level inserted
+///     `is_update`:  whether the operation is an update operation
+/// # Returns
+///     `[Bool;6]`: the next state
+fn next_state(
```
katat, line 145: it'd better to have more documentation on the logic for each transition.

src/stdlib/native/smt/lib.no
```diff
@@ -0,0 +1,446 @@
...
+
+fn assert_vstate_valid(state:[Bool;5]) {
+    let mut sum = 0;
+    for ii in 0..5 {
+        sum = sum + get_val(state[ii]);
+    }
+    assert_eq(sum , 1);
+}
+
+fn v_next_state(
```
katat, line 352: need some documentation here

src/stdlib/native/smt/lib.no
```diff
@@ -0,0 +1,446 @@
...
+        roots = new_roots;
+    }
+
+
+    // if there is delete then switch the old root and new root
+    let is_delete = operation[0] && operation[1];
+    roots = switcher(is_delete , roots[0] , roots[1]);
+
+    let is_ins_del_upd = get_val(enabled);
+    let new_root = (is_ins_del_upd * (roots[1] - old_root)) + old_root;
```
katat, line 309: it will be more explicit to use if else here.

src/stdlib/native/smt/lib.no
```diff
@@ -0,0 +1,446 @@
...
+
+
+    let old_leaf = compute_leaf_hash(old_key,old_value);
+    let new_leaf = compute_leaf_hash(new_key,new_value);
+   
+    let old_key_bits = bits::to_bits(LEN , old_key);
+    let new_key_bits = bits::to_bits(LEN , new_key);
+
+    let mut level_inserted = old_level_inserted(enabled,siblings);
+    let mut states = [[false;6]; LEN];
```
katat, line 263: would it be easier if it is a field value to represent the different states intead of using an array of booleans?

src/stdlib/native/smt/lib.no
```diff
@@ -0,0 +1,446 @@
...
+}
+
+/// switches left and right inputs based on the insert sel bool 
+/// if self bool is true we switch l and r else same
+fn switcher(sel: Bool,left: Field, right: Field) -> [Field;2] {
+    let sel_val = get_val(sel);
+    let aux = sel_val * (left - right);  
+    let new_left = left - aux;
+    let new_right = right + aux;
+    return [new_left,new_right];
```
katat, line 96: can we just use if else to switch based on bool?

katat (commented):
Nice work!
I just had a round of review, and will need another rounds. I left some comments.

I think some of the arithmetic logic can be simplified using ifelse block. 
More documentation on the arithmetic logic in the `compute_roots` and `next_state`, if they can't be simplified `ifelse`, would be useful.

src/stdlib/native/smt/lib.no
```diff
@@ -0,0 +1,446 @@
...
+///
+/// # Parameters
+/// `key`: key for the node
+/// `value`:  Value for the node
+///
+/// # Returns
+/// Field: mimc7 hash for key = 1 and values = [key,value]
+fn compute_leaf_hash(key: Field, value: Field) -> Field {
+    let hash_values = [key, value];
+    return mimc::mimc7_hash(hash_values, 1);
```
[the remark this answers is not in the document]
0xnullifier: Yeah this was in the iden3 specification I think this is mainly for domain seperation of the internal and leaf hash functions though I cannot find a reason why

src/stdlib/native/smt/lib.no
0xnullifier: Got it

src/stdlib/native/smt/lib.no
0xnullifier: yeah but that makes the states computation a bit more difficult as in this case I can directly do boolean operations as my state transitions are dictated by booleans. 
Also the assertion that there is only one state at a time

I tried the single state variable earlier but it got too difficult but when we do #248 it will be way easier maybe we can refactor then?

One more point is that `circomlib` also uses 5 varibales for this exact reason maybe it has less constraints this way?

src/stdlib/native/smt/lib.no
0xnullifier: yeah sure I just ran a quick test and both have same number of constraints so I will switch to the bool method. I just did this way cause I was too focused on `circomlib` methods they do this type of pattern everywhere for some reason

src/stdlib/native/smt/lib.no
```diff
@@ -0,0 +1,446 @@
...
+// state       
+// top             old_root                                        new_root
+//                 /      \                                       /        \
+// top        [sibling]   [<val>]                            [sibling]    [<update_val>]
+//                         /      \                                         /      \ 
+// new                [sibling]   [old_key] <-                         [sibling]  [<new-val>] --> this will change in case of update or del
+// The key we are doing         -----------                                       /    \ 
+// operation follows this path and this key will be                          [old_key] [new-key]
+
+// verfication is more simpler we just compute roots using the siblings and keys , values
```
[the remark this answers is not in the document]
0xnullifier: that's just trash english what i meant to say is that verification is more simpler process we just need to compute roots using siblings , keys and value

src/stdlib/native/smt/lib.no
```diff
@@ -0,0 +1,446 @@
...
+///     - `Field` The new computed root
+fn compute_new_root(
+    old_root: Field,
+    old_key: Field,
+    old_value: Field,
+    new_key: Field,
+    new_value: Field,
+    siblings: [Field; LEN], // the number of siblings is equal to the number of hash
+    isold_zero : Bool, // indicates that we are inserting in a empty leaf
+    operation: [Bool; 2]
```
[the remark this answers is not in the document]
0xnullifier: Yeah sure we can do that just makes the more sense

src/stdlib/native/smt/lib.no
0xnullifier: this is not needed anymore as operation is now a single variable

src/stdlib/native/smt/lib.no
```diff
@@ -0,0 +1,434 @@
...
+){
+
+    // for non inclusion proofs and is_nf_zero false `not_found_key` and `key` should not be equal
+    let valid_keys = (inclusion_proof || (not_found_key != key)) && !is_nf_zero;
+    assert(valid_keys);
+
+    let not_found_leaf = compute_leaf_hash(not_found_key,not_found_val);
+    let leaf = compute_leaf_hash(key,value);
+
+    let nfk_bits = bits::to_bits(LEN , not_found_key);
```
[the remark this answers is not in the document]
katat: maybe add a comment here to note this is just for range checking.

katat (commented):
Sorry for the delay. 
Finally got a chance to go through the code logic :D
Lgtm, @mimoo might want to have another round.

I am wondering if we can encapsulate the code a bit more using structs. The idea is like:
```rust
struct SMT {
    root: Field,
    ...
}

// init smt with a root
fn SMT.new(root, ...) -> SMT {
...
}

// differentiate the operations via different methods
// once updated, it should update the SMT.root internally
fn SMT.insert(...)
fn SMT.update(...)
fn SMT.delete(...)

// for verifying the membership
fn SMT.verify_membership(...)
fn SMT.verify_non_membership(...)
```

For the non membership verification logic, I am wondering what are the assumptions.

There seems no comparisons in terms of branch navigations among the following two keys in order to indicate the `key` doesn't exist because of the proof of exists of the `not_found_key`. 

Is this something agreed externally in practice, or actually something missing in the circuit? In other word, how can it prove the `key` doesn't exist by proving its relationship with `not_found_key`, which has been proven exist already.

0xnullifier:
Hey @katat the run fails for some reason it was working when I tested it before but it seems that there is a bug in the way we are typechecking in the mast for `ExprKind::IfElse`. 
https://github.com/zksecurity/noname/blob/a9a83b8d8cfa14f7391a3ed840908a09e14e931e/src/mast/mod.rs#L1066
As my code fails due to unmatching types if `Field {const : true}` and `Field { const: false}` if I change it to something like this :
```rust
            let is_match = match (&then_mono.typ, &else_mono.typ) {
                (Some(then_ty), Some(else_ty)) => then_ty.match_expected(else_ty, false),
                // handle all the case
            };
```
Then it works

0xnullifier:
> I am wondering if we can encapsulate the code a bit more using structs. The idea is like:

This looks better I will refactor the code to something like this and push it

src/stdlib/native/smt/lib.no
```diff
@@ -0,0 +1,434 @@
...
+    let mut outputs = [0;2];
+    outputs[0] = if sel { right } else { left };
+    outputs[1] = if sel { left } else { right };
+    return outputs;
+}
+
+/// This function get the level where the key will be inserted. level 0 is the root , 1 is next and so on..
+/// To find this level all the child level must have a sibling of 0 and
+/// the parent level has a sibling != 0. The root is assumed to have a parent level with a sibling != 0
+fn level_inserted(siblings: [Field;LEN]) ->  [Bool;LEN] {
```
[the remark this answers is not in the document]
0xnullifier: yup there is no need for the full array but then I would have to slice the array which would be expensive so I think it is better to do this correct me if I'm wrong

katat:
> As my code fails due to unmatching types if Field {const : true} and Field { const: false} if I change it to something like this :

Oh yeah, nice spot!
Do you want to create a separate PR to fix this?

0xnullifier:
> Oh yeah, nice spot!
Do you want to create a separate PR to fix this?

Do I just raise the pr directly or open a issue and then raise a pr?

> but how does this relate to proving not_found_leaf occupies where the membership key is supposed to be in the path?

I think i don't get the question. The `compute_root` function place this root in place of where key is supposed to be like. So here is what I am saying
image
To compute the green node we need the `not_found_key` right ? If we did not have that we would not have any way of computing the root and checking if the given root is equal to the computed root. Is there anything wrong with this logic?

katat:
> Do I just raise the pr directly or open a issue and then raise a pr?

It is fine to just create a pr and point it to the comment above for reference.

> I think i don't get the question. The compute_root function place this root in place of where key is supposed to be like. So here is what I am saying

Nice diagram, thanks!
I think I got it now. The `compute_root` always takes the key bits to determine the branches. So if the `not_found_leaf` is proven exist via the `key_bits`, it implies the `key` doesn't exist. 

```
let new_root = compute_root(
    states[(LEN - 1) - idx],
    computed_root,
    siblings[(LEN - 1) - idx],
    key_bits[(LEN - 1) - idx], // the branching bit is always based on the key
    not_found_leaf,
    leaf
);
```

PR #250, opened 2025-01-04
fix: invalid folded mutable constant value

katat:
When mutable constant variable is modified in a for loop, it should be marked as a non constant field. 
Otherwise the MAST phase will try to constant fold that variable into wrong constant value, because we don't support unrolling in for loop.

### DIFF
--- a/src/negative_tests.rs
+++ b/src/negative_tests.rs
@@ -473,6 +473,29 @@ fn test_generic_custom_type_mismatched() {
 }
 
 #[test]
+fn test_generic_mutated_cst_var_in_loop() {
+    let code = r#"
+        fn gen(const LEN: Field) -> [Field; LEN] {
+            return [0; LEN];
+        }
+        
+        fn main(pub xx: Field) {
+            let mut loopvar = 1;
+            for ii in 0..3 {
+                loopvar = loopvar + 1;
+            }
+            let arr = gen(loopvar);
+        }
+        "#;
+
+    let res = tast_pass(code).0;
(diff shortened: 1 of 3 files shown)


src/negative_tests.rs
```diff
@@ -472,6 +472,29 @@ fn test_generic_custom_type_mismatched() {
...
+        fn gen(const LEN: Field) -> [Field; LEN] {
+            return [0; LEN];
+        }
+        
+        fn main(pub xx: Field) {
+            let mut loopvar = 1;
+            for ii in 0..3 {
+                loopvar = loopvar + 1;
+            }
+            let arr = gen(loopvar);
```
mimoo, line 487: I'm not sure I understand why this code shouldn't compile

katat: Because we don't support for loop unrolling, so the `loopvar` won't end up with the correct constant value during MAST. 

With the wrong constant value for the `loopvar`, the generic function will be instantiated with wrong generic parameter.

mimoo: what would it take to make this code work though? I feel like we should support simple constant folding like this

mimoo (approved):
maybe let's create an issue to fix that? I don't think we should prevent this kind of code :o

src/negative_tests.rs
katat: as long as the `gen(loopvar)` is not accepting the `loopvar` as its const generic parameter, it would work.

PR #260, opened 2025-01-07
allow expressions in ifelse block

katat:
- Allows the expressions in `ifelse` block.
- Allow constant field and non constant field in branches.

Related to https://github.com/zksecurity/noname/issues/248


mimoo:
so let's close this one due to our convo?

PR #261, opened 2025-01-11
Adds tuple type 

### LINKED ISSUE #255: Support tuple return type
Tuple return type can be useful when it needs to return more than one type from a function, avoiding the need to duplicate functions into two versions just for different return types.

For example:

```rust
fn divmod_limbs(const BITLEN: Field, dividend: [Field; DVDLEN], divisor: [Field; DVSLEN]) 
-> ([Field; (DVDLEN - DVSLEN) + 1], [Field; DVSLEN]) {
    let quotient_limbs = unsafe calculate_quotient_limbs(BITLEN, dividend, divisor);
    let rem_limbs = unsafe calculate_rem_limbs(BITLEN, dividend, divisor);
    
    ...

    return (quotient_limbs, rem_limbs);
}
```

Without the support of tuple type, this function has to be duplicated with the only difference in the return type.

0xnullifier:
closes #255 

the example file can be found here [tuple.no](https://github.com/0xnullifier/noname/blob/feat-tuple/examples/tuple.no) this works and runs the output is:
image

I have not implemented destructuring as that is potentially another pr.

The code is pretty much identical to array type as tuple is just a multi-type array

and also as @mimoo suggested that we should keep the tuple access similar to array access. I have just changes the code for `ExprKind::ArrayAccess` to `ExprKind::ArrayOrTupleAccess` as they are identical at lexer level and dealt with this at type checker and writer.

it also refactors the array logging and combines it to array or tuple logging

### DIFF
--- a/examples/tuple.no
+++ b/examples/tuple.no
@@ -1,4 +1,3 @@
-// tuples as field for structs
 struct Thing {
     xx: Field,
     tuple_field: (Field,Bool)
@@ -15,24 +14,34 @@ fn Thing.new(xx: Field , tup: (Field,Bool)) -> (Thing , (Field,Bool)) {
     );
 }
 
+fn generic_array_tuple_test(var : ([[Field;NN];LEN],Bool)) -> (Field , [Field;NN]) {
+    let zero = 0;
+    let result = if var[1] {var[0][LEN - 1][NN - 1]} else { var[0][LEN - 2][NN - 2] };
+    return (result , var[0][LEN - 1]);
+}
+
 // xx should be 0
-fn main(pub xx: Field) -> (Field , (Thing,Bool)){
-    // creation of new tupple with different types
+fn main(pub xx: [Field; 2]) -> Field {
+    // creation of new tuple with different types
     let tup = (1, true);
 
     // create nested tuples 
-    let nested_tup = (1, (true , [1,2,3]));
+    let nested_tup = ((false, [1,2,3]), 1);
     log(nested_tup); // (1, (true , [1,2,3]))
-    // access nested tuples
-    let incr = nested_tup[1][1][0]; // 1
+    
+    let incr = nested_tup[1]; // 1
 
     // tuples can be input to function
-    let mut thing = Thing.new(xx , (xx , xx == 0));
+    let mut thing = Thing.new(xx[1] , (xx[0] , xx[0] == 0));
 
     // you can access a tuple type just like you access a array 
     thing[0].tuple_field[0] += incr;
-    
+    log(thing[0].tuple_field[0]);
+    let new_allocation = [xx,xx];
+    let ret = generic_array_tuple_test((new_allocation, true));
+
     assert_eq(thing[0].tuple_field[0] , 1);
+    log(ret[1]); // logs xx i.e [0,123]
     
-    return (xx ,(thing[0],thing[0].tuple_field[1]));
+    return ret[0];
 }
--- a/examples/tuple.no
+++ b/examples/tuple.no
@@ -1,15 +1,38 @@
-
+// tuples as field for structs
 struct Thing {
     xx: Field,
-    tuple_field: (Field,Field)
+    tuple_field: (Field,Bool)
 }
 
-fn main(pub xx: Field) -> (Field , Thing){
-    // let some = dummy_fn(1,2);
-    let mut tup = (1, true);
-    let mut thing = Thing {xx : xx , tuple_field: (xx , xx+1)};
-    tup[1] = false;
-    log(tup[0]);
-    log(tup[1]);
-    return (xx ,thing);
+// return tuples from functions
+fn Thing.new(xx: Field , tup: (Field,Bool)) -> (Thing , (Field,Bool)) {
+    return (
+        Thing {
+        xx: xx,
+        tuple_field:tup
+        }, 
+        tup
+    );
+}
+
+// xx should be 0
+fn main(pub xx: Field) -> (Field , (Thing,Bool)){
+    // creation of new tupple with different types
+    let tup = (1, true);
+
+    // create nested tuples 
+    let nested_tup = (1, (true , [1,2,3]));
+    log(nested_tup); // (1, (true , [1,2,3]))
+    // access nested tuples
+    let incr = nested_tup[1][1][0]; // 1
+
+    // tuples can be input to function
+    let mut thing = Thing.new(xx , (xx , xx == 0));
+
+    // you can access a tuple type just like you access a array 
+    thing[0].tuple_field[0] += incr;
+    
+    assert_eq(thing[0].tuple_field[0] , 1);
+    
+    return (xx ,(thing[0],thing[0].tuple_field[1]));
 }
--- a/examples/tuple.no
+++ b/examples/tuple.no
@@ -1,0 +1,13 @@
+
+
+struct Thing {
+    xx: Field,
+    tuple_field: (Field,Field)
+}
+
+fn main(pub xx: Field) -> (Field , Thing){
+    // let some = dummy_fn(1,2);
+    // let tup = (1, true);
+    let thing = Thing {xx : xx , tuple_field: (xx , xx+1)};
+    return (xx ,thing);
+}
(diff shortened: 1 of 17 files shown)


0xnullifier:
we can also do a better example but I cannot just think of any

0xnullifier:
hey @katat can you have a look thanks!

src/type_checker/checker.rs
```diff
@@ -484,6 +487,10 @@ impl<B: Backend> TypeChecker<B> {
                 let el_typ = match typ.typ {
                     TyKind::Array(typkind, _) => *typkind,
                     TyKind::GenericSizedArray(typkind, _) => *typkind,
+                    TyKind::Tuple(typs) => match &idx.kind {
+                        ExprKind::BigUInt(index) => typs[index.to_usize().unwrap()].clone(),
```
katat, line 491: This should return an error `TupleIndexOutofBounds` if the index is out of bound.

src/circuit_writer/ir.rs
```diff
@@ -987,28 +987,39 @@ impl<B: Backend> IRWriter<B> {
...
                             return Err(self.error(
                                 ErrorKind::ArrayIndexOutOfBounds(idx, *array_len as usize - 1),
                                 expr.span,
                             ));
                         }
-                        ty
+                        (**ty).clone()
+                    }
+                    TyKind::Tuple(typs) => {
+                        if idx >= typs.len() {
```
katat, line 1006: out of bound check should already done during type checking phase.

src/circuit_writer/writer.rs
```diff
@@ -742,31 +751,46 @@ impl<B: Backend> CircuitWriter<B> {
...
                             ));
                         }
-                        ty
+                        let ele_ty = (**ty).clone();
+                        let start = idx * self.size_of(&ele_ty);
+                        (start, ele_ty)
+                    }
+
+                    TyKind::Tuple(typs) => {
+                        if idx >= typs.len() {
```
katat, line 777: out of bound check should already done during type checking phase.

src/mast/mod.rs
```diff
@@ -1137,6 +1146,30 @@ fn monomorphize_expr<B: Backend>(
                 return Err(error(ErrorKind::InvalidArraySize, expr.span));
             }
         }
+
+        ExprKind::TupleDeclaration(items) => {
+            // forcing for the same size as the max size for array
```
katat, line 1151: do you mean `checking` instead of `forcing` ?

examples/tuple.no
```diff
@@ -0,0 +1,38 @@
+// tuples as field for structs
+struct Thing {
+    xx: Field,
+    tuple_field: (Field,Bool)
+}
+
+// return tuples from functions
+fn Thing.new(xx: Field , tup: (Field,Bool)) -> (Thing , (Field,Bool)) {
```
katat, line 8: Does it support generic array types such that the following signature works?
```
fn Thing.new(xx: Field, tup: ([Field; LEN],Bool)) -> (Thing, (Field, [Bool; LEN]))
```

src/circuit_writer/writer.rs
```diff
@@ -204,7 +204,7 @@ impl<B: Backend> CircuitWriter<B> {
                             ))?,
                         };
 
-                        // compute the size of each element in the array
+                        // compute the size of each element in thearray
```
katat, line 207: format error

src/circuit_writer/writer.rs
```diff
@@ -742,31 +751,46 @@ impl<B: Backend> CircuitWriter<B> {
...
+                let array_typ = self
+                    .expr_type(container)
+                    .expect("cannot find type of container");
 
-                let elem_type = match array_typ {
+                // real starting index for narrowing the var depends on the cotainer
+                // for arrays the indexing for the vars is just the idx*len(typ) but for
+                // tuples we will have different types thus start = idx as nested tuples are treated as
+                // `Vec<TyKind>` where for nested arrays the start becomes the idx * [each_array_size]
+                let (start, elem_type) = match array_typ {
```
katat, line 763: might be better to return type length instead of `elem_typ`

src/circuit_writer/writer.rs
```diff
@@ -742,31 +751,46 @@ impl<B: Backend> CircuitWriter<B> {
...
+                    }
+
+                    TyKind::Tuple(typs) => {
+                        if idx >= typs.len() {
+                            return Err(self.error(
+                                ErrorKind::TupleIndexOutofBounds(idx, typs.len()),
+                                expr.span,
+                            ));
+                        }
+                        (idx, typs[idx].clone())
```
katat, line 783: The start index `idx` doesn't look correct. `idx` here is the current element type index of the tuple, not the start index of the cell variable inside an element type in the tuple.

I think the correct way to calculate the start index is:
1. get all the tuple element types before this `idx`.
2. sum up all the size of these element types.
3. the sum is the start index.

src/circuit_writer/ir.rs
```diff
@@ -987,28 +987,39 @@ impl<B: Backend> IRWriter<B> {
...
-                        ErrorKind::UnexpectedError("expected array"),
+                        ErrorKind::UnexpectedError("expected container"),
                         expr.span,
                     ))?,
                 };
 
-                // compute the size of each element in the array
-                let len = self.size_of(elem_type);
+                // compute the size of each element in the container
+                let len = self.size_of(&elem_type);
```
katat, line 1022: Same issue I mentioned in the circuit writer.
The start index for the tuple doesn't look correct.

katat:
Looks good. I left you some comment.

The key issue is the way of calculating the starting index to narrow the cell variables for an tuple element in the circuit writer and IR writer.

katat (commented):
Looks good. I left you some comment.

The key issue is the way of calculating the starting index to narrow the cell variables for an tuple element in the circuit writer and IR writer.

Also I would suggest having a separate expression type `TupleAccess`, instead of merging both the array access and tuple access into `ArrayOrTupleAccess`. Having them separated is to avoid the potential assumption confusions such as the calculation of the start index, although some of the codes are duplicated, which is fine.

examples/tuple.no
0xnullifier: will have to check but it should as it is just `Vec<TyKind>` and everything is recursive

src/mast/mod.rs
0xnullifier: yup checking

0xnullifier:
> Also I would suggest having a separate expression type TupleAccess, instead of merging both the array access and tuple access into ArrayOrTupleAccess.

@katat  how can you go about doing this cause at the parser they are exactly same and only at the type checker can we distinguish that whether it was ArrayAccess or TupleAccess.


Is there any way for me to find the type of Variable we access in the parser like here 
```rust

  // array or tuple access
  Some(Token {
      kind: TokenKind::LeftBracket,
      ..
  }) => {
     // do some type checking here ? to find whether self is tuple variable or not

      Expr::new(
          ctx,
          ExprKind::ArrayAccess {
              container: Box::new(self),
              idx: Box::new(idx),
          },
          span,
      )
  }
```

One way I can do this is define a  `ExprKind::Access` here and then at the typechecker mutate the variable but this would result in a mutable expr in the `compute_type` function.

katat:
Yeah, indeed the parser doesn't have enough information to know whether it is array or tuple. Changing the Expr type might complicate the flow also. 

So looks like we have to keep `ArrayOrTupleAccess `.

src/circuit_writer/writer.rs
0xnullifier: yup this makes more sense after playing around I realised the my way runs because I was just lucky the way example was made. Fixed this

nohup.out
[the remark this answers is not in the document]
0xnullifier: oh  mb did not see it in the diff for some reason

src/mast/mod.rs
```diff
@@ -1148,7 +1164,7 @@ fn monomorphize_expr<B: Backend>(
         }
 
         ExprKind::TupleDeclaration(items) => {
-            // forcing for the same size as the max size for array
+            // checking if the size of tuple same is as maximum size of an array
```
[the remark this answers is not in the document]
0xnullifier: sure got it

0xnullifier:
done!

PR #264, opened 2025-01-22
bug-fix: correct type matching in if-else expression

0xnullifier:
found this bug while implementing smt in stdlib [here](https://github.com/zksecurity/noname/pull/249#issuecomment-2604436495)


katat (approved):
thanks!

PR #266, opened 2025-01-25
Hint ite statements

0xnullifier:
@katat I think I could not explain things in the issue so I have raised a draft pull request. For the example of `if_else.no` taken in the pr it works as the blocks early returns but if I change the example to something like this:
```rust
hint fn ite(xx: Field , arr: [Field;LEN]) -> Field {
    let mut var = 0;
    if xx == 10 {
        var = xx;
        for idx in 0..LEN {
            var = var + arr[idx];
        }
        var = xx * var;
    } else {
       var = xx * xx;
    }
    return var;
}
```
this causes the then_branch return type to panic. So how can I support these type of statements? moreover what about only if statements?

### DIFF
--- a/src/circuit_writer/ir.rs
+++ b/src/circuit_writer/ir.rs
@@ -507,25 +507,27 @@ impl<B: Backend> IRWriter<B> {
                 else_branch,
             } => {
                 fn_env.nest();
-                let cond_var = self.compute_expr(fn_env, condition)?;
-                let cond_val = self
+                let cond_var = self
                     .compute_expr(fn_env, condition)?
                     .unwrap()
                     .value(self, fn_env)
                     .cvars[0]
                     .clone();
-                let ret = self.compile_block(fn_env, then_branch)?;
+                let ret = self.compile_block(fn_env, then_branch)?.unwrap().cvars[0].clone();
+                let mut else_ret = None;
                 match else_branch {
                     Some(stmts) => {
                         //start another block
                         fn_env.nest();
-                        self.compile_block(fn_env, stmts)?;
+                        else_ret = self.compile_block(fn_env, stmts)?;
                         fn_env.pop();
                     }
                     None => (),
                 }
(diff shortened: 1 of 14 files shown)


src/circuit_writer/ir.rs
```diff
@@ -501,6 +501,34 @@ impl<B: Backend> IRWriter<B> {
...
+                else_branch,
+            } => {
+                fn_env.nest();
+                let cond_var = self
+                    .compute_expr(fn_env, condition)?
+                    .unwrap()
+                    .value(self, fn_env)
+                    .cvars[0]
+                    .clone();
+                let ret = self.compile_block(fn_env, then_branch)?.unwrap().cvars[0].clone();
```
katat, line 516: why then_branch block isn't handled the same way as the else_branch block?

katat: shouldn't the fn_env.nest() and fn_env.pop() be called for each `compile_block`?

katat (commented):
The reason it panics is the example you provided returns none for the ITE branches.

Let's first agree on the grammar on this enhanced ITE statement.

I would suggest to still align with the grammar supported in the constraint code:

```
let res = if cond {t1} else {t2};
```

In hint function for now, we would just add the support of return statement. 

In your example, we would instead enforce it to be:

```
hint fn ite(xx: Field , arr: [Field;LEN]) -> Field {
    let mut var = 0;
    let res = if xx == 10 {
        var = xx;
        for idx in 0..LEN {
            var = var + arr[idx];
        }
        xx * var // align with the existing grammar
    } else {
       return xx * xx // or it returns
    }
    return res;
}
```

This way aligns with the existing grammar in constraint mode, while avoiding adding too much complexity for this enhancement at this stage.

0xnullifier:
@katat hey sorry for the delay I am a little busy with some work this week. I will pick this up again after the weekend

PR #270, opened 2025-04-24
ci: bump actions/checkout to v4

dizer-ti:
Node 20 compatibility: switch all jobs to actions/checkout@v4. Pure maintenance, behavior unchanged.


mimoo (approved):
thx!

PR #275, opened 2025-06-15
Update documentation link to the current Noname book URL

GarmashAlex:
Replaced the outdated link to the Noname book (mimoo.github.io/noname) with the new official URL (zksecurity.github.io/noname) in the crate-level documentation comment. This ensures users are directed to the latest and maintained documentation for the Noname language.


mimoo (approved):
thx so much!

PR #288, opened 2025-09-30
Add feature gates 

winderica:
This PR adds feature gates for `cli`, `r1cs`, `kimchi`, and `server`. It allows downstream libraries to select desired features instead of enabling all of them.

The version of arkworks libraries is also bumped to the latest.


mimoo:
can I ask you why you're doing this or looking into noname?

winderica:
Hey @mimoo! I am opening this PR because recently I tried to bump the version of dependencies to latest in Sonobe, which has experimental support for DSLs including Noname, but I faced some [compilation issues](https://github.com/privacy-ethereum/sonobe/actions/runs/18101346347/job/51504842333) when targeting wasm.

Then I realized that Noname provides some functionalities that are useful as a CLI tool, and for this purpose it depends on wasm-unfriendly libraries such as `tokio`. On the other hand, Sonobe (and maybe some other downstream projects) imports Noname only for constraint synthesis/witness extraction. To allow for both use cases, I added feature gates to Noname, so that normal users can continue to enjoy all functionalities of Noname as a binary, while projects that use Noname as a library can disable unused features.

src/utils/mod.rs
```diff
@@ -82,6 +84,74 @@ pub fn title(res: &mut String, s: &str) {
     writeln!(res).unwrap();
 }
 
+/// Field element helpers
+///   Unless otherwise stated everything is in little-endian byte order.
+pub trait FieldHelpers<F> {
```
mimoo, line 89: these have moved here https://github.com/o1-labs/proof-systems/blob/c1803e0a480172c193ba67c826110980303a172f/utils/src/field_helpers.rs#L72 so no need to copy it in noname

src/tests/mod.rs
```diff
@@ -1,3 +1,5 @@
+#[cfg(feature = "kimchi")]
 mod examples;
```
mimoo, line 2: feels weird to gate this under kimchi, but I guess maybe less painful than add the feature gate within examples?

